PR #39, opened 2018-04-04
Add more details into the README

mostlygeek:
Add info about schedule and proposed API for RemoteSettings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,8 @@ const records = await RemoteSettings("my-key").get();
   ]
 */
 
-for(const entry of records) {
-  // Do something with entry, e.g:
-  // await InternalAPI.load(entry.id, entry.label, entry.weight);
+for(const record of records) {
+  // await InternalAPI.load(record.id, record.label, record.weight);
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -14,3 +14,48 @@ These attributes differentiate V1 from V2:
 ## When will it be available? 
 
 * We are targetting Firefox 62 (nightly) to have the new APIs available
+
+## What will the API Look like? 
+
+There are two new APIs to work with settings:
+
+1. `get()` 
+2. `on()` 
+
+### get()
+
+```js
+const { RemoteSettings } = ChromeUtils.import("resource://services-common/remote-settings.js", {});
+
+const records = await RemoteSettings("my-key").get();
+
+/*
+  records == [
+    {label: "Yahoo",  enabled: true,  weight: 10, id: "d0782d8d", last_modified: 1522764475905},
+    {label: "Google", enabled: true,  weight: 20, id: "8883955f", last_modified: 1521539068414},
+    {label: "Ecosia", enabled: false, weight: 5,  id: "337c865d", last_modified: 1520527480321},
+  ]
+*/
+
+for(const entry of records) {
+  // Do something with entry, e.g:
+  // await InternalAPI.load(entry.id, entry.label, entry.weight);
+});
+```
+
+### on() 
+
+The `on()` allows handlers to be triggered when records changes on the server side. 
+
+```js
+RemoteSettings("my-key").on("create", (event) => { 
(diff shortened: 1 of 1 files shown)


mostlygeek:
@leplatrem going through the [gist](https://gist.github.com/leplatrem/a17eb395b6aba3834df1b5033358bf22) I wasn't sure how `.on("change", ... )` is different from `.on('update', ...)`.  It isn't clear what the difference is between these two.

leplatrem:
> how .on("change", ... ) is different from .on('update', ...).

Yes, good point. 

Currently we have `on("change", handler)` which is triggered whenever something changes on the server side. 

But since we have the ability to distinguish entries that were created, updated or deleted, I thought we could introduce to fine grained events. They all work the same way, the event has a `data` attribute with the list. I can't find a better event name than `update` for entries that were modified :| (`replace`, `modify`, `amend`, ...)

Other options:
- rename `change` to `sync` (which is closer to reality BTW)
- have only one `change` event but the list of created/updated/deleted as in different attributes
- rename to past tense? created/updated/deleted?

README.md
```diff
@@ -14,3 +14,48 @@ These attributes differentiate V1 from V2:
...
+
+1. `get()` 
+2. `on()` 
+
+### get()
+
+```js
+const { RemoteSettings } = ChromeUtils.import("resource://services-common/remote-settings.js", {});
+
+const records = await RemoteSettings("my-key").get();
```
leplatrem: `records` was intentional here? Maybe `entries` to be consistent with the `for` loop?

leplatrem:
I think the approach I prefer would be to have: `sync`, `create`, `update`, `delete`, all with `data`

README.md
mostlygeek: it was intentional. I thought it would be clearer. I don't mind changing it to `entries`

mostlygeek:
I think I would prefer `sync`, `create`, `update` and `delete` as well. 

So what does `on("sync", ...)` give me?  It sounds like it gives the list of _created_ and _updated_ entries. Does it have _deleted_ entries? If so, I don't know how that would work ...

leplatrem:
> So what does on("sync", ...) give me?

I would say the whole new list in a `data` attribute.

We just have to make a decision, whether we have distinct events all with a `data` attribute that contains the event related set of entries (`e.data`), or a single event with several attributes (`e.all`, `e.created`, `e.updated`, `e.deleted` ....)

mostlygeek:
I think we should just have 

```js
on("sync", (e) => {
   // e.updated = [ ... ]
   // e.created = [ ... ]
   // e.deleted = [ ... ] 
});
```

as the only thing we support in v2.  It's less API surface and documentation.  If we need to in future versions we can add more event types. For example: `on("error", ...)`.

mostlygeek:
@leplatrem what do you think about slimming the listen down to just `on("sync", ...)`?  We're bike shedding a bit but it's good to get this clear :)

leplatrem:
The proposition to have a single event with several attributes seems acceptable. 
And I think that in most cases we'll need the whole list too. Of course we could do:

```js
.on("sync", async (e) => {
  const client = e.target;
  const records = await client.get();
  // ...
})
```
But that's a pity to let every listener refetch it, because we already have it at hand in the piece of code that sends the event.

What about?

```js
on("sync", (e) => {
   // e.data.all = [ ... ]
   // e.data.updated = [ ... ]
   // e.data.created = [ ... ]
   // e.data.deleted = [ ... ] 
});
```

leplatrem:
And using destructuring it could just be written like this:
```js
on("sync", ({ data: { all } }) => {
   // all.forEach(...)
});
```

mostlygeek:
Had discussion this is the `on()` API: 

```js
on("sync", (e) => {
   // e.data.current = [ ... ]
   // e.data.updated = [ ... ]
   // e.data.created = [ ... ]
   // e.data.deleted = [ ... ] 
});
```

PR #76, opened 2018-07-10
Serve attachments in the local container (fixes #75)

### LINKED ISSUE #75: Tutorial is missing setup for serving attachments files
Attachments are stored in `/tmp` inside the container. But, nothing serves them in HTTP.

leplatrem:
fixes #75

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+attachments/
+docs/_build/
+server.ini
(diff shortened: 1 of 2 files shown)


conman124:
This worked great.  The only thing I would add is that I needed to change the permissions on the attachments folder to be world writable.  I'm not sure if this was strictly necessary, but without it, the kinto server gave this error: `PermissionError: [Errno 13] Permission denied: '/tmp/attachments/main/public-suffix-list/c7235747-169f-4969-afb1-fd95481a3ad3.bin'`

PR #104, opened 2019-10-01
Add tutorial for local development of Kinto-Admin with signoff

leplatrem:


docs/tutorial-dev-kinto-admin.rst
```diff
@@ -0,0 +1,156 @@
...
+         -H 'Content-Type:application/json' \
+         -u editor:3d1t0r
+
+
+    curl -X PATCH --fail --verbose $SERVER/buckets/main-workspace/groups/password-recipes-reviewers \
+         -H 'Content-Type:application/json-patch+json' \
+         -d '[{ "op": "add", "path": "/data/members/0", "value": "account:reviewer" }]' \
+         -u $BASIC_AUTH
+
+And at last, create some records, request review and approve changes.
```
glasserc: This almost looks like literate code. Maybe we should just put all the code in a script and ship the script? Of course, then I would want to write the script in Python.

docs/tutorial-dev-kinto-admin.rst
```diff
@@ -0,0 +1,156 @@
...
+
+    npm install
+    npm start
+
+The UI should be accessible at http://0.0.0.0:3000
+
+Initialization script
+---------------------
+
+Since the container is not configured with a real database by default, the content is flushed on each restart.
```
glasserc: Maybe add something like `This means you will have to populate data regularly.`

docs/tutorial-dev-kinto-admin.rst
leplatrem: The whole repo relies on `curl`.

And I thought it would make more sense to explain what we were doing here, instead of just `wget http://.../init.sh | bash`

PR #105, opened 2019-12-26
Add reference to how to edit STAGE/PROD manifest to add users

englehardt:
Someone (like me) trying to figure out how to manage user permissions for a collection is likely to go to the multi-signoff tutorial. This adds a small comment at the top to redirect them to the "getting started" page.


englehardt:
> This link is technically internal, right? Is it necessary to provide a hyperlink here or is there some smart thing we can do with Sphinx?

I initially didn't see a way to link to a sub-section of a doc, but it looks like it's possible. See [the docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations). I've updated the reference to that.

PR #110, opened 2020-05-05
Remote Settings Dev Tools is the canonical way to switch environments

### LINKED ISSUE #108: Discourage setting preferences manually
It's common for people to try to just set the Remote Settings server URL without setting any of the other associated preferences. One possible source of this error is in https://remote-settings.readthedocs.io/en/latest/tutorial-dev-server.html#prepare-the-client. Let's tighten up the wording a little bit to make it clear that Remote Settings dev tools is the "official" way to change environment, perhaps not listing the preferences at all.

glasserc:
Fixes #108.

I wasn't really sure what to do about the `dev` environment. As far as I can tell, it's impossible to use correctly, so maybe we should just delete the whole article?

### DIFF
--- a/docs/tutorial-dev-server.rst
+++ b/docs/tutorial-dev-server.rst
@@ -22,7 +22,12 @@ This guide assumes you have already installed and set up the following:
 Introduction
 ------------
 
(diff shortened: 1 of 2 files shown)


leplatrem:
:heart:

PR #114, opened 2020-05-07
improving wording about target filters to make their purpose clearer

### LINKED ISSUE #111: Is JEXL filtering an anti-pattern?
Coming out of the Experiments work week, targeting/filtering is quite hard to get right.  Should we deprecate/remove JEXL filtering from remote-settings?

mostlygeek:
I think this is all we need to close #111.  

During the work week it wasn't clear what the intended purpose of target filters is.  To allow for a single server side collection to serve many a small subset of user groups. 

An unintended consequence that occurred is that target filtering seems suitable for experimentation.  To be suitable we would need to maintain parity with the experiment platform's (normandy, normandy dev tools, filter objects, capabilities, suitabilities, etc) targeting feature set. 

Maintaining parity would diverge from the original purpose and create additional engineering overhead. 

This doc update is meant to make clear what the purpose and ideal usage of target filtering is.

### DIFF
--- a/docs/target-filters.rst
+++ b/docs/target-filters.rst
@@ -3,14 +3,12 @@
 Target filters
 ==============
 
-By default, all records are made available to all users.
+By default all records in a collection are made available.  However, there are some use cases where it is more practical to use a single server side collection and filter record visibility in the browser.  Target filters allows for these use cases.
 
-When defined on a particular record, target filters describe which users should have it.
-
-Basically, filters are conditional expressions executed locally in the client's browser and, if they pass, the corresponding record is available. Filters have access to information about the user, such as their locale, addons, and Firefox version.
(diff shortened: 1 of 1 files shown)


docs/target-filters.rst
```diff
@@ -3,14 +3,15 @@
 Target filters
 ==============
 
-By default, all records are made available to all users.
+By default all records in a collection are made available.  However, there are some use cases where it is more practical to use a single server side collection and filter record visibility in the browser.  Target filters allows for use cases.
```
leplatrem, line 6: *allows for use cases* ? Is this a typo or a grammar construction I don't know?

mostlygeek: Thank you.  Fixed.

PR #118, opened 2020-11-23
Update synchronization information

mostlygeek:
Updated some out of date documentation on when Remote settings synchronizes with the server.


leplatrem (approved):
:+1:

PR #121, opened 2021-01-15
Fix link to VPN

tublitzed:
I was poking around at the docs, and noticed we have a link [on this page](https://remote-settings.readthedocs.io/en/latest/getting-started.html) pointing to a moved wiki page. Figured I'd fix while I'm in here.


leplatrem (approved):
Thank you for taking the time to fix it!

PR #124, opened 2021-02-11
Fix broken bucket links on the getting started doc

TGiles:
This fixes the two broken links referencing how to view records of a particular collection


leplatrem (approved):
Oh! Good catch! Thank you :100: :pray:

PR #142, opened 2022-01-24
Remove superfluous files and config

leplatrem:
By removing the custom Autograph config, we also align our integration tests with the official way to verify signatures.

The certificate chains, created inside Autograph containers on startup, are exposed via a simple httpd server.


grahamalama (approved):
🧹

PR #154, opened 2022-01-25
Ship default configuration (fixes #151)

### LINKED ISSUE #151: Default configuration and automatic setup on empty database
Projects integrating with Remote Settings want to have integration tests. Running a local remote settings instance should be super simple.

Currently, our consumers setup servers that look more like a Kinto instance than Remote Settings. For example, they push the records to the `main` bucket directly to setup their tests. This differs from our DEV/STAGE/PROD environments. In order to unify the experience of a local container with our environments requires a lot of setup (config and objects creation) that we don't want to duplicate in every consumer repo. 

We want to provide a Remote Settings docker container that works out of the box.

One idea would ship a config file for tests, and the container be ran with `KINTO_INI=/etc/testing.ini`

With this config, the container would:
- use the memory storage backend
- initialize the buckets (eg. `main-workspace`)
- be permissionless (a specific account like `admin:s3cr3t` can do everything, or everything open to everyone)
- use a local signer (maybe we don't want #1948 in the end)
- no multisignoff enabled (users can approve their own changes)

----

imported from https://github.com/mozilla-services/kinto-dist/issues/1963

leplatrem:
Fixes #151 

* [x] figure out why uwsgistart complains about port
* [x] Read `KINTO_SIGNER_RESOURCES` from env, to be able to have the server auto create a specific collection on startup --> #156 
* [ ] Proof read tutorial

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,7 @@ COPY --from=compile /dogstatsd_plugin.so .
 
 ENV PYTHONUNBUFFERED=1 \
     PORT=8888 \
+    KINTO_INI=/etc/kinto.ini \
     PATH="/opt/venv/bin:$PATH"
 
 # add a non-privileged user for installing and running
@@ -55,6 +56,9 @@ RUN chown 10001:10001 /app && \
 
 COPY . .
 RUN pip install ./kinto-remote-settings
+
+# Use the example config as default.
+RUN mv config/example.ini /etc/kinto.ini
 
 # Generate local key pair to simplify running without Autograph out of the box (see `config/testing.ini`)
 RUN python -m kinto_remote_settings.signer.generate_keypair /app/ecdsa.private.pem /app/ecdsa.public.pem
@@ -68,4 +72,4 @@ EXPOSE $PORT
 
 # Run uwsgi by default
 ENTRYPOINT ["/bin/bash", "/app/bin/run.sh"]
-CMD ["uwsgi", "--ini", "/etc/kinto.ini"]
+CMD uwsgi --ini ${KINTO_INI}
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,9 @@ RUN chown 10001:10001 /app && \
 COPY . .
 RUN pip install ./kinto-remote-settings
 
+# Generate local key pair to simplify running without Autograph out of the box (see `config/testing.ini`)
+RUN python -m kinto_remote_settings.signer.generate_keypair /app/ecdsa.private.pem /app/ecdsa.public.pem
+
 # Drop down to unprivileged user
 RUN chown -R 10001:10001 /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,7 +45,7 @@ COPY --from=compile /dogstatsd_plugin.so .
 
 ENV PYTHONUNBUFFERED=1 \
     PORT=8888 \
-    KINTO_INI=/etc/kinto.ini \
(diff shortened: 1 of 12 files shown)


docs/tutorial-local-server.rst
```diff
@@ -221,40 +95,17 @@ Both containers should be connected, and the heartbeat endpoint should only retu
...
-         -d '{"permissions": {"read": ["system.Everyone"]}}' \
-         -H 'Content-Type:application/json' \
-         -u $BASIC_AUTH
+.. note::
 
-    curl -X PUT ${SERVER}/buckets/main \
-         -d '{"permissions": {"read": ["system.Everyone"]}}' \
-         -H 'Content-Type:application/json' \
-         -u $BASIC_AUTH
+    Another option is to use ``docker-compose`` with a cloned repository of https://github.com/mozilla/remote-settings
```
grahamalama, line 106: Is it worth documenting that this is run with:
```
$ make build
$ docker-compose up
```

Dockerfile
```diff
@@ -45,6 +45,7 @@ COPY --from=compile /dogstatsd_plugin.so .
 
 ENV PYTHONUNBUFFERED=1 \
     PORT=8888 \
+    KINTO_INI=/etc/kinto.ini \
```
grahamalama, line 48: Any reason we don't want to set this to `config/example.ini`? Then we don't need the `mv` command below.

bin/run.sh
```diff
@@ -18,7 +16,7 @@ case $1 in
     kinto start --ini $KINTO_INI
     ;;
   uwsgistart)
-    KINTO_INI=$KINTO_INI uwsgi --http :8888 --ini $KINTO_INI
+    uwsgi --http :$PORT --ini $KINTO_INI
```
grahamalama, line 19: Should we set a default `PORT`? This is set in the `Dockerfile`, so most of the time we're covered, but in case someone is running this service without Docker, this script does make the assumption that there's a `PORT` env var set.

Similarly, why did we remove the default `KINTO_INI` env var? If we don't want to provide a default, can we display a message reminding the user to set that env var if one isn't set?

docs/tutorial-local-server.rst
leplatrem: Isn't `make build` supposed to be the same as `docker-compose build`?

bin/run.sh
leplatrem: I like the fact that all environment variables are listed and defaulted in the same place. Either Dockerfile or run.sh, but not both.

Also, using the `run.sh` outside the Docker context isn't something that we advocate anywhere, so I don't think we should worry about this.

docs/tutorial-local-server.rst
grahamalama: Not quite. If I just run `docker-compose build` in a fresh environment, I get this error:
```
 => ERROR [kinto:tests internal] load metadata for docker.io/library/remotesettings:build    
```

That's because `Dockerfile.Testing` starts 
```Dockerfile
FROM remotesettings:build
```

And since that container hasn't been built yet and it's not available via Dockerhub, it throws the error.

So, `make build` builds `remotesettings:build` first so that the testing container can be built on top of it. (It also goes ahead and builds the test container.)

We could alternatively instruct people to run the `web` container specifically with `docker-compose up web`.

leplatrem: Gotcha. Will update the docs

PR #155, opened 2022-01-26
Remove ability to configure group names (fixes #149)

### LINKED ISSUE #149: Remove configurable group name
Now that kinto-signer is remote-settings, we can remove useless settings, like for example the ability to configure the group names at various levels.

We could just have a default value set to:
```
kinto.signer.editors_group = {collection_id}-editors
kinto.signer.reviewers_group = {collection_id}-reviewers
```

**Caution**: we may have to maintain some metadata in the capabilities, since it's used by Kinto Admin. See https://github.com/Kinto/kinto-admin/blob/e5792ea176192d42d61086588f9622e40876ee9b/src/plugins/signoff/components.tsx#L22-L38

-----

Imported from https://github.com/mozilla-services/kinto-dist/issues/1945

leplatrem:
fixes #149

### DIFF
--- a/kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
+++ b/kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
@@ -53,10 +53,6 @@ def pick_resource_and_signer(request, resources, bucket_id, collection_id):
     return resource, signer
 
 
-def resource_group(resource, group):
-    return group.format(collection_id=resource["source"]["collection"])
-
-
 def sign_collection_data(event, resources, **kwargs):
     """
     Listen to resource change events, to check if a new signature is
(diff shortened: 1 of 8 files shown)


kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
```diff
@@ -60,10 +53,7 @@ def pick_resource_and_signer(request, resources, bucket_id, collection_id):
     return resource, signer
 
 
-def resource_group(resource, name, default):
-    group = resource.get(name, default)
-    # If review is configured per-bucket, the group patterns have to be replaced
-    # with the source collection id.
+def resource_group(resource, group):
```
grahamalama, line 56: Do you think it's still worth it to have this `resource_group` function when now all we're doing is calling `<group>.format`?

PR #157, opened 2022-01-26
Make sure we don't check groups when review is disabled

leplatrem:


kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
```diff
@@ -308,7 +308,7 @@ def check_collection_status(
 
         # 2. work-in-progress -> to-review
         elif new_status == STATUS.TO_REVIEW:
-            if editors_group_uri not in user_principals:
+            if editors_group_uri not in user_principals and _to_review_enabled:
```
grahamalama, line 311: Should we add a separate guard clause to with a message that reviews are disabled?

kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
```diff
@@ -320,7 +320,7 @@ def check_collection_status(
                 raise_invalid(message="Collection already signed")
 
             # Only allow to-sign from to-review if reviewer and no-editor
-            if reviewers_group_uri not in user_principals:
+            if reviewers_group_uri not in user_principals and _to_review_enabled:
```
grahamalama, line 323: Similar comment as above -- I noticed all of the guard clauses in this branch contain `and _to_review_enabled`. Would it make sense to make `if _to_review_enabled` its own guard with a message stating that review is disabled?

kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
leplatrem: I think we don't want that. This way when working on client scripts, they would work in all environments, including DEV where review is disabled

kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
leplatrem: Done!

PR #161, opened 2022-01-28
Only run build_and_publish on main branch or git tag

grahamalama:
Based on the filters set for `build_and_publish`, prior to this commit,
I would have expected this job to only run on tagged commits. However,
this job seemed to run on every commit, leading me to beleive the regex
provided would match an empty tag. So, the regex was changed to match a
non-empty string.

Additionaly, a filter was added so that the build_and_publish job will
only run on the "main" branch. After this commit is merged, we expect
this job to run on tagged commits or after any commit to the "main"
branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,4 +162,7 @@ workflows:
             - integration_test
           filters:
             tags:
-              only: /.*/
+              only: /.+/
+            branches:
+              only: main
+              


leplatrem (approved):
I think it was done this way to test the container build on each test. But since then we run the tests from the container, so we're good

grahamalama:
@leplatrem I'm still not 100% confident about the before and after of this commit 🤔.

I think after looking more at the Circle [documentation](https://circleci.com/docs/2.0/workflows/#using-contexts-and-filtering-in-your-workflows) for filtering workflows:

Before: run `build_and_publish` on branches AND all tagged commits. I think this essentially meant we would run `build_and_publish` on every push. And I think 

> prior to this commit, I would have expected this job to only run on tagged commits

was an incorrect assumption 


After: run `build_and_publish` only on commits pushed to main, and on any tagged commit

But I'm a little confused by the 2nd example in [this section of the docs](https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag):

```yaml
workflows:
  version: 2
  build-n-deploy:
    jobs:
      - build:
          filters:  # required since `deploy` has tag filters AND requires `build`
            tags:
              only: /.*/
      - deploy:
          requires:
            - build
          filters:
            tags:
              only: /^v.*/
            branches:
              ignore: /.*/
```

Specifically, the comment on the filters tag of the build job. Do we need to specify the tags filter in the jobs that `build_and_publish` requires?

Issue #152, opened 2022-01-24
Run integration tests on any server 


leplatrem:
Follow-up of mozilla-services/kinto-dist#1927  

We should make sure (and document) how to run the integration on any server providing the appropriate environment variables and setup.

For example, in our DEV/STAGE server we could have a dedicated user (`account:integration-test`) and collection (`product-integrity`), and run the full integration suite. 

The suite would skip bucket, collection, and user creation, clear all the collection records, and execute the whole test suite.

This would replace the current https://github.com/mozilla-services/kinto-integration-tests/ which is basically useless

--- 

imported from https://github.com/mozilla-services/kinto-dist/issues/1965

---

Remaining tasks, as of March 6, 2023:
- [x] https://github.com/mozilla/remote-settings/pull/227
- [x] https://github.com/mozilla/remote-settings/issues/377

PR #162, opened 2022-01-28
Use `local.ini` as default config for tests

leplatrem:
Currently we have two example config files that serve the same role `example.ini` and `local.ini`. This pull-request will clean this up and merge the two into `local.ini`.

The final `local.ini` in this PR is very close to what is deployed in DEV/STAGE, and since integration tests run against it, it will also pave the way for #152 

The integration test about monitor/changes had to be rewritten, because it was watching the `main-workspace` bucket, which does not reflect our typical deployment (watch main and main-preview buckets)


leplatrem:
This PR is currently blocked by #166  but review by commit should be doable

config/local.ini
```diff
@@ -122,5 +152,18 @@ args = (sys.stderr,)
 level = NOTSET
 formatter = color
 
+[handler_sentry]
+class = raven.handlers.logging.SentryHandler
+args = ('https://<key>:<secret>@app.getsentry.com/<project>',)
+level = ERROR
+formatter = generic
```
grahamalama, line 159: Is this something we want enabled locally? Or does it not really matter if someone runs the container locally with this enabled?

grahamalama (approved):

config/local.ini
leplatrem: I hesitated, but thought that it wouldn't harm to leave it.  Alternatively we could have a `prod.ini` file example with that sort of things

PR #176, opened 2022-02-08
Remove hardcoded bucket and collection names

leplatrem:


requirements-dev.txt
```diff
@@ -5,7 +5,7 @@ canonicaljson-rs
 flake8
 httpie
 isort
-kinto-http
```
grahamalama, line 8: Any reason we want to have >=10.9.0 for this dependency? Doesn't a lack of version number mean "latest" implicitly?

grahamalama (approved):

requirements-dev.txt
leplatrem: The bucket and collection attributes on the client were added in 10.9.0, that's why. But I'm ok with latest implicitly

PR #181, opened 2022-02-17
Modify testing Docker, docker-compose setup

grahamalama:
We used to have the Dockerfiles for the server and tests in the same
directory. This is because we built the tests container on top of the
server, and our build setup required the containers to share the same
context.

Now, we build the containers with independent contexts. As a result, we
can move the testing Dockerfile to the tests directory. The only
additional changes needed to complete this migration were to:
- split out a testing-specific requirements.txt and include it in that
  directory
- add a named volume for database data so that in wouldn't be
  recreated during the tests after running the migrations


docker-compose.yml
```diff
@@ -1,8 +1,12 @@
 version: "3"
 
+volumes:
+  db-data:
 services:
   db:
     image: postgres:12
+    volumes:
+      - db-data:/var/lib/postgresql/data
```
grahamalama, line 9: I'm still not _entirely_ sure why we have to include this volume now when we didn't before.

What I _think_ is happening is that when we run `docker-compose run tests`, it re-creates a separate `web` container with a new data volume for the command. This new container doesn't have the migrations applied. By adding this shared volume, we "persist" the migration between commands. 

But again, even if this is the case, I'm not sure why we weren't running into this issue before.

leplatrem: This makes sense indeed. I'm wondering whether we should have a `.volumes/` folder with all of them (certs chains, etc.)

PR #182, opened 2022-02-17
Circle config updates - docs depdency cache and build_and_publish job

grahamalama:
- Add separate cache for docs deps
- Remove build_and_publish job from pr workflow
  - since we only want to run this job for merges to main and tagged commits


leplatrem (approved):
Thanks!

Issue #186, opened 2022-03-02
CDN caching issue after signature refresh 


leplatrem:
Users reported certificate expiration errors, from two different contexts (Merino Rust client and Desktop Firefox).

> pjenvey> FYI, this morning merino began reporting certificate expired errors when verifying RS signatures
> pjenvey> it's using the rust remote settings client. almost wondered if it was a false alarm but I noticed some talk of expired certs recently
>

> jschanck> I'm seeing invalid signatures for a few collections when using the remote-settings-devtools extension. Hitting "clear local data" followed by "force sync" on main/message-groups gives me InvalidSignatureError: Invalid content signature (main/message-groups) . Same for main/personality-provider-models
>
> jschanck> I dumped the cert chain here: https://searchfox.org/mozilla-central/source/services/settings/RemoteSettingsClient.jsm#861 and it is using the [expired one](https://content-signature-2.cdn.mozilla.net/chains/remote-settings.content-signature.mozilla.org-2022-03-01-08-35-12.chain) rather than the one listed in the [collection metadata](https://content-signature-2.cdn.mozilla.net/chains/remote-settings.content-signature.mozilla.org-2022-04-10-18-14-48.chain).

The error could not be reproduced the following day. 

This could indicate a CDN caching issue.

When synchronization happens from a push notification, the current data timestamp is passed in the `?_expected=` query param, in order to bust the CDN cache.

However there are several situations where the synchronization happens without push notification, and in these cases the current data timestamp is taken from the [monitor/changes](https://firefox.settings.services.mozilla.com/v1/buckets/monitor/collections/changes/records) endpoint. 

Regularly we refresh the collections signature, to make sure the metadata always point to the latest certificate chain. However, this operation doesn't «bump» the data timestamp. 

The following scenario seems to be the culprit:

1. changes are made on collection X, data timestamp is bumped (`timestamp=42`)
2. clients pull data from push notification, changeset is [aggressively cached](https://github.com/mozilla/remote-settings/blob/d1996490b2f5dc7f4d098c14f89dc86f3f568ca1/kinto-remote-settings/src/kinto_remote_settings/changes/views.py#L126-L138) on the CDN with `_expected=42` (for [1 hour](https://github.com/mozilla-services/cloudops-deployment/blob/b2ed06d77b6d84860b7bada0834c866fe37b289f/projects/kinto/puppet/modules/kinto/templates/kinto.ini.erb#L144))
[...]

PR #187, opened 2022-03-03
Add metadata timestamp to monitor/changes entries (ref #186)

leplatrem:


docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+1. Use metadata timestamp as cache bust
+1. Use metadata timestamp as cache bust only if signature fails
+
+## Decision Outcome
+
+Chosen option: option X because XXXX
+
+Both options 1 and 2 have a significant impact.
+
+Option 1 will cause more trafic, but globally absorbed by the CDN. It will force reverification of up-to-date data on clients, solving [bug 1640126](https://bugzilla.mozilla.org/show_bug.cgi?id=1640126). Rolling out is immediate and affect all versions.
```
grahamalama, line 91:
```suggestion
Option 1 will cause more traffic, but globally absorbed by the CDN. It will force reverification of up-to-date data on clients, solving [bug 1640126](https://bugzilla.mozilla.org/show_bug.cgi?id=1640126). Rolling out is immediate and affect all versions.
```

docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+
+Option 2 will double trafic on the origins, but have no impact on clients. Rollout will take time.
```
grahamalama, line 93:
```suggestion
Option 2 will double traffic on the origins, but have no impact on clients. Rollout will take time.
```

docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+
+**Impact**
+
+High.
+
+Bumping the data timestamp on signature refresh will have a significant impact both on servers and on clients.
+
+We refresh the signature of all collections that haven't been updated in the last 7 days. With this option, we will force synchronization on all clients for these collections every at most 7 days.
+
+This will increase trafic on our servers, although the impact will be relatively low on the origins.
```
grahamalama, line 144:
```suggestion
This will increase traffic on our servers, although the impact will be relatively low on the origins.
```

docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+
+The clients would use this `metadata_timestamp` field instead of `last_modified` as the value passed in the `?_expected=` query param to bust the cache.
+
+**Complexity**
+
+Low. The overall architecture would not be modified.
+
+**Impact**
+
+High. This will **multiply by 2 the trafic on our origin servers**, since we would have a population of clients that would bust the cache with one timestamp value, and the rest with another value.
```
grahamalama, line 197:
```suggestion
High. This will **multiply by 2 the traffic on our origin servers**, since we would have a population of clients that would bust the cache with one timestamp value, and the rest with another value.
```

docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+
+This increase of trafic is likely to remain forever, given that legacy clients will keep on polling changes for a very long time. And only one distinct request on the CDN is enough to lead to a request on the origin server.
```
grahamalama, line 199:
```suggestion
This increase of traffic is likely to remain forever, given that legacy clients will keep on polling changes for a very long time. And only one distinct request on the CDN is enough to lead to a request on the origin server.
```

docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+- **Complexity**: Low → High: how much additional complexity
+- **Impact**: Low → High: how big is the change on the whole service
+- **Efforts**: Low → High: how much efforts are necessary
+- **Rollout**: Slow → Fast: how fast can we roll out the fix
+
+## Considered Options
+
+1. Bump data timestamp on signature refresh
+1. Use metadata timestamp as cache bust
+1. Use metadata timestamp as cache bust only if signature fails
```
grahamalama, line 83:
```suggestion
0. [Option 0 - Do nothing](#option-0---do-nothing)
1. [Option 1 - Bump data timestamp on signature refresh](#option-1---bump-data-timestamp-on-signature-refresh)
1. [Option 2 - Use metadata timestamp as cache bust](#option-2---use-metadata-timestamp-as-cache-bust)
1. [Option 3 - Use metadata timestamp as cache bust only if signature fails](#option-3---use-metadata-timestamp-as-cache-bust-only-if-signature-fails)
```

docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Decision Outcome
+
+Chosen option: option X because XXXX
```
grahamalama, line 87: Have we landed on a decision yet? Or are we still waiting on SRE input?

docs/adr/adr_001_cache_bust.md
```diff
@@ -0,0 +1,322 @@
...
+
+This will increase trafic on our servers, although the impact will be relatively low on the origins.
+
+> Note: traffic on origin shouldn't be too affected. And we are very likely to refresh signatures of several collections are once, therefore grouping them in one push notification.
+
+During synchronization, the clients will apply the empty list of changes (delta since last sync), and revalidate the signature of their local data.
+
+This will have an impact on clients CPUs on every signature refresh. But this will indirectly fix [bug 1640126](https://bugzilla.mozilla.org/show_bug.cgi?id=1640126), which consists in verifying local data when up-to-date.
+
+> Note: We could adjust the [behaviour of the lambda](https://github.com/mozilla-services/remote-settings-lambdas/blob/443117cb16cc9d720c59341c6c811e7bf127b289/commands/refresh_signature.py#L78-L93) to control the number of refreshed collections in one run.
```
grahamalama, line 152: How much of an impact do you think this would have? Would it mitigate some of the points you made above such that it would drop the impact to a medium or a low?

grahamalama (commented):
Mostly fixing a typo with the word `traffic` 😅, but also a few other questions.

docs/adr/adr_001_cache_bust.md
leplatrem: I'd like to hear inputs from SRE yes. 

Also, I realized we don't have a native way to bump the data timestamp on a collection from the `kinto.core.storage` API. We would have to add it for this specific use-case, which is not great given this API hasn't changed in 6 years :D

docs/adr/adr_001_cache_bust.md
leplatrem: We have like 20-30 collections that weren't updated in the last 6 months. It's hard to tell how many are actually instantiated in clients, and whose signature refresh would imply initiating a synchronization. But if we keep the threshold of refreshing every 7 days at most, we could refresh 4-5 collections a day, which would definitely have a low impact on clients.

PR #199, opened 2022-03-31
Restructure integration tests

grahamalama:
Split the one, long integration tests file into separate files, one for each "concern". This will make iterating on the tests easier.

git / Github thinks I renamed `tests/integration_test.py → tests/integration/plugins/test_signer.py` since the file content is very similar, but it was an add/delete I did in separate commits. I'm going to rebase both of these commits when I merge the PR so that history is preserved.


leplatrem (approved):
:+1:

PR #200, opened 2022-04-06
Build containers once, load in tests and deploy jobs

grahamalama:
Currently, we build the Remote Settings server and test containers in 3 different jobs:
- `integration_test`
- `browser_test`
- `build_and_publish`

Now, we build the containers once in a `build_containers` job, persist the containers to a [workspace](https://circleci.com/docs/2.0/workspaces/), then load containers in the jobs that need them.


leplatrem (approved):
LGTM

PR #204, opened 2022-04-12
Fix `400 Bad request` when creating collection in batch requests (fixes #203)

### LINKED ISSUE #203: Collection creation issue from batch requests
**Steps to reproduce**

- Collection `a` exists
- A batch request is sent, in which we create a collection `b` and update the collection `a`
```
POST /batch
{
    "requests": [
        {
            "method": "PUT",
            "path": "/buckets/main-workspace/collections/b",
        },
        {
            "method": "PATCH",
            "path": "/buckets/main-workspace/collections/a",
            "body": {"data": {"foo": 42}},
        },
    ],
}
```

**Expected result**

- The server creates the collection `b`
- The server updates the colllection `a`

**Actual result**

- The server responds `400 Invalid Request`, `Cannot set status to 'signed'`

**Explanations**

- When a collection is created, we receive a `ResourceChanged` event, and we use an event listener to sign it right away ([here](https://github.com/mozilla/remote-settings/blob/14f9baeb366bcb6b2e9ae9fee252d0f39a68d91a/kinto-remote-settings/src/kinto_remote_settings/signer/__init__.py#L243-L248))
- When we sign a collection, we update the source attributes to set its `status` to `signed`. In order to keep track of this change, we emit a `ResourceChanged` event. We call this a *cascade* event, since it's sent from an event listener.
- When several events are sent for the same action and resource name, we merge them together. There will be one event, but with several `impacted_objects`. 

- In the batch request example above, when we process the collection `b` creation, and its initial signing, a `ResourceChanged` event is sent for the `status` switch to `signed`. This event gets merged with the `ResourceChanged` event emitted for the PATCH update of collection `a`.
- The issuer of the event for the PATCH update is a normal user (author of the batch request). However, the one for the signed status event is the signer code itself (`user id = plugin:signer`)
- When events get merged, we «loose» some information about the merged one, including its issuer.
- The check that we have in place to make sure that no user can manually set a collection status to `signed` ignores update events that were issued by the signer code. But since the *signer event* was merged into the *user event*, the check code raises an invalid request `Cannot set status to 'signed'`, assuming the status change was performed by a user and not by the signer code.

**Possible solutions**

- Prevent event merging when their issuer differs
- Prevent event merging of cascade events with upstream events

leplatrem:
fixes #203


grahamalama (approved):

grahamalama:
I guess with this PR, we can close https://github.com/mozilla/remote-settings/pull/209 when this gets merged 🙌🏻

PR #221, opened 2022-05-06
Push integration test container to Dockerhub

grahamalama:
In addition to pushing the Remote Settings server container to https://hub.docker.com/u/mozilla/remote-settings, we now push the integration test container to https://hub.docker.com/r/mozilla/remote-settings-integration-tests.

### DIFF
--- a/bin/deploy-dockerhub.sh
+++ b/bin/deploy-dockerhub.sh
@@ -21,6 +21,20 @@ function retry() {
     return 0
 }
 
+# Usage: push_container_to_repo CONTAINER REPO TAG
+# Push container to specified repo with tag
+# Example: push_container_to_repo remotesettings mozilla/remote-settings latest
+function push_container_to_repo() {
+    container=$1
+    repo=$2
+    tag=$3
+    docker tag $container "$repo:$tag" ||
+        (echo "Couldn't tag $container as $repo:$tag" && false)
+    retry 3 docker push "$repo:$tag" ||
+        (echo "Couldn't push $repo:$tag" && false)
+    return 0
+}
+
 # configure docker creds
 echo "$DOCKER_PASS" | docker login -u="$DOCKER_USER" --password-stdin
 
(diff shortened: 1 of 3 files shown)


bin/deploy-dockerhub.sh
```diff
@@ -26,13 +26,27 @@ echo "$DOCKER_PASS" | docker login -u="$DOCKER_USER" --password-stdin
...
+    docker tag $SERVER_CONTAINER "$SERVER_DOCKER_REPO:$TAG" ||
+        (echo "Couldn't tag $SERVER_CONTAINER as $SERVER_DOCKER_REPO:$TAG" && false)
+    retry 3 docker push "$SERVER_DOCKER_REPO:$TAG" ||
+        (echo "Couldn't push $SERVER_DOCKER_REPO:$TAG" && false)
+
+    docker tag $TEST_CONTAINER "$INTEGRATION_TEST_DOCKER_REPO:$TAG" ||
+        (echo "Couldn't tag $TEST_CONTAINER as $INTEGRATION_TEST_DOCKER_REPO:$TAG" && false)
+    retry 3 docker push "$INTEGRATION_TEST_DOCKER_REPO:$TAG" ||
+        (echo "Couldn't push $INTEGRATION_TEST_DOCKER_REPO:$TAG" && false)
+
```
leplatrem, line 50: I have the feeling that we could have a bash function instead of duplicating everything

leplatrem (approved):
Should we document how to run these tests in the README? In a follow-up?

bin/deploy-dockerhub.sh
grahamalama: Good call. Added here d3554ffc55db98f6b785394c7af6e6d4423ca5cf.

grahamalama:
> Should we document how to run these tests in the README? In a follow-up?

Yeah I'd say documenting in a followup is a good call. This container will really only be for CI/CD though, right? Users can always `docker-compose run tests` (and supply different environment variables as needed).

bin/deploy-dockerhub.sh
```diff
@@ -21,6 +21,20 @@ function retry() {
     return 0
 }
 
+# Usage: push_container_to_repo CONTAINER REPO TAG
+# Push container to specified repo with tag
```
leplatrem, line 25: nit: I would invert these two lines

bin/deploy-dockerhub.sh
```diff
@@ -21,6 +21,20 @@ function retry() {
...
+# Example: push_container_to_repo remotesettings mozilla/remote-settings latest
+function push_container_to_repo() {
+    container=$1
+    repo=$2
+    tag=$3
+    docker tag $container "$repo:$tag" ||
+        (echo "Couldn't tag $container as $repo:$tag" && false)
```
leplatrem, line 32: This works, and is equivalent to what we had before.

if we want to improve that, we should exit if this fail instead of just echoing. Something like:

```code
docker tag $container "$repo:$tag"
res=$?
if [ $res -ne 0 ] ; then
    echo "Couldn't tag $container as $repo:$tag"
    return $res
fi
```
But that may be a bit zealous

leplatrem (approved):

bin/deploy-dockerhub.sh
grahamalama: Yeah, maybe we can look through this file in another PR and fix things like this -- I mostly just tried to keep things consistent with what we were already doing in this PR.

PR #235, opened 2022-06-10
Integration test changes 

grahamalama:
Make changes to the integration test suite so that it will pass on the dev server


tests/integration/plugins/test_signer.py
```diff
@@ -263,6 +264,7 @@ async def test_signer_plugin_reviewer_verifications(
 
     await upload_records(editor_client, 5)
 
+    # TODO: this doesn't raise. Why not?
```
grahamalama, line 267: I'm guessing this has something to do with some permission configured for the dev instance but not for the local instance.

tests/conftest.py
```diff
@@ -176,6 +179,10 @@ async def flush_default_collection(
     setup_auth: Auth,
 ):
     yield
+    # TODO: we should still set "setup_auth" to delete collection? So, edit
+    # readme to suggest setting SETUP_AUTH whether the tests are being run on
+    # DEV or locally? Or could EDITOR_AUTH be used here? Who is allowed to
+    # delete these records?
```
leplatrem, line 185: SETUP_AUTH has admin perms. It can create and delete collections.

EDITOR_AUTH is a remote settings user.

For LOCAL, we setup the server. 
On DEV we shouldn't do any operation that require admin perms, and SETUP_AUTH can be omitted

tests/integration/plugins/test_signer.py
```diff
@@ -46,7 +46,8 @@ async def test_signer_plugin_capabilities(make_client: ClientFactory):
     anonymous_client = make_client(tuple())
     capability = (await anonymous_client.server_info())["capabilities"]["signer"]
     assert capability["group_check_enabled"]
-    assert capability["to_review_enabled"]
+    # TODO: this is disabled on dev
+    # assert capability["to_review_enabled"]
```
leplatrem, line 50: I would suggest that we remove this assertion here, but use the `capability["to_review_enabled"]` to adjust the scope of the `test_signer_plugin_full_workflow` below (or skip it completely if it's too coupled with review enabled)

tests/integration/plugins/test_signer.py
leplatrem: Exactly

leplatrem:
You may want to use this https://github.com/Kinto/kinto-http.py/pull/282

tests/conftest.py
grahamalama: > On DEV we shouldn't do any operation that require admin perms, and SETUP_AUTH can be omitted

But do we still want to "clean up" after ourselves on DEV? Otherwise, we'll accumulate more and more records after every integration test. 

I think the `EDITOR_AUTH` user creates all of the records in our integration tests -- can they also delete their own records? If so, maybe we can use their credentials for `flush_default_collection`.

leplatrem: > I think the EDITOR_AUTH user creates all of the records in our integration tests -- can they also delete their own records? If so, maybe we can use their credentials for flush_default_collection.

Indeed, an editor can delete their own records. But since editor is among the users that have write permissions on the `product-integrity` collection, they can also delete every records in it.

tests/integration/plugins/test_signer.py
grahamalama: Any ideas on what we should do about this? Should we skip some of these assertions if we're running on DEV?

leplatrem: What about:
- inspect the `to_review_enabled` for the collection in the root URL capabilities
- skip certain assertions (if that doesn't bloat the test too much) if the flag is False
?

tests/integration/plugins/test_changes.py
```diff
@@ -19,24 +20,29 @@ async def test_changes_plugin(
...
+    test_collection_records = [
+        record for record in records if record["collection"] == source_collection
+    ]
+    assert len(test_collection_records) == 2
 
     editor_client = make_client(editor_auth)
     resource = await signed_resource(editor_client)
+    assert "bucket" in test_collection_records[0]
+    assert test_collection_records[0]["bucket"] == resource["destination"]["bucket"]
+    assert test_collection_records[1]["bucket"] == resource["preview"]["bucket"]
```
grahamalama, line 34: These assertions are pretty flaky. Sometimes these assertions pass as-is, sometimes switching to

```python
    assert test_collection_records[0]["bucket"] == resource["preview"]["bucket"]
    assert test_collection_records[1]["bucket"] == resource["destination"]["bucket"]
```

is the thing that works.

@leplatrem any ideas here? Should this order actually matter? I also kind of don't know how 2 records are magically present in the collection before we upload any.

leplatrem: In theory, the list of latest changes is sorted by `-last_modified`, and `destination` should be at `[0]`, and `preview` at `[1]`.
Because when we approve changes, we first copy the records to `preview`,  and then to `destination`.

If that is the only blocker, we could rewrite the assertion to remove notion of order and open an issue to fix the tests later.

tests/conftest.py
```diff
@@ -173,11 +188,10 @@ def _make_client(auth: Auth) -> AsyncClient:
 @pytest_asyncio.fixture(autouse=True)
 async def flush_default_collection(
     make_client: ClientFactory,
-    setup_auth: Auth,
+    editor_auth: Auth,
 ):
     yield
-    setup_client = make_client(setup_auth)
-
+    setup_client = make_client(editor_auth)
```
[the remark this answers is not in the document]
grahamalama: https://github.com/mozilla/remote-settings/pull/235/commits/8fdb48bc307527ccbdd12e18492222210de6aa8a

tests/conftest.py
```diff
@@ -85,6 +86,12 @@ def pytest_addoption(parser):
         default=DEFAULT_SKIP_SERVER_SETUP,
         help="Skip server setup operations",
     )
+    parser.addoption(
+        "--to-review-enabled",
+        action="store_true",
+        default=DEFAULT_TO_REVIEW_ENABLED,
+        help="Include tests and related to the to-review config option",
+    )
```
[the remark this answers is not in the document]
grahamalama: I think having this as a CLI flag (or somehow not relying on introspection alone) might be useful. For example, we expect `to-review` to be disabled on the dev server. Say it was somehow enabled. If we used introspection to determine which signer test to run, it would pick the "right" one, but it would break our "assertion" that `to-review` was disabled.
What are your thoughts there?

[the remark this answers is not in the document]
leplatrem: I agree, I hadn't looked at it this way.

If we want to check that the server has a particular setting, then we have to store the expected value somewhere.
Either, we introspect and do ugly stuff like `"DEV" in server_info["name"] and to_review is False` or we pass it in CLI like you did.

If we want to be smarter here (with rules, or default values by ENV etc.) then we can still improve it over time.

tldr; ignore my comment and leave the CLI flag.

leplatrem (approved):

tests/integration/plugins/test_changes.py
grahamalama: 👍🏻 [24b589f](https://github.com/mozilla/remote-settings/pull/235/commits/24b589fc3ca9f462211014149ecbe7797d36ed01)

https://github.com/mozilla/remote-settings/issues/377

leplatrem (approved):
:shipit:

PR #236, opened 2022-06-11
Fix circle CI issues

grahamalama:
Docker container builds were failing in CircleCI and had been for some time. The builds were failing because UWSGI was failing to build. 

It seems that the change that fixed the builds was to set a `VIRTUAL_ENV` environment variable in the Docker container that points to the virtual environment we create. I'm not exactly sure _why_ this change fixes things, however.

In addition to this change, we also made a few more small improvements WRT Docker and CircleCI:
- cache `.venv` in CircleCI
- don't install system and Python packages that seemed to have no effect on the images building
- update the base image of the server and test Docker containers from Python `3.10.4` to `3.10.5`
- use a non-slim image for the server "compile" stage since we need to install a compiler to build a few dependencies anyway
- update the Ubuntu machine runner to the latest version (`ubuntu-2004:2022.04.2`)
- set a language attribute in Sphinx conf (this was throwing an error when compiling the docs)
- Add more entries to the `.dockerignore`
- run pip commands with a `python -m` prefix to ensure we're using our virtual environment (and not the system pip)


leplatrem (approved):

leplatrem:
Awesome Graham! Thanks for your patience on this one 💯

PR #237, opened 2022-06-16
Install Rust in docker and cache python dependencies by version

leplatrem:
This PR caches the python dependencies by Python version.

Without this, if the version of Python is upgraded, the cached compiled dependencies can contain symbols and headers of the previous one. Leading to weird compilation errors (namely with uwsgi).

> Note, this is a follow-up of #236 

This PR also adds Rust to the container in order to compile `python-canonicaljson-rs`, in case the wheels published on Pypi are not available for the host platform (eg. arm).
See https://github.com/mozilla-services/python-canonicaljson-rs/issues/3 and https://github.com/mozilla-services/python-canonicaljson-rs/pull/8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,11 +8,11 @@ version: 2.1
 aliases:
   - &restore_deps_cache
     name: Restoring Python dependency cache
-    key: v3-requirements-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
+    key: v3-requirements-{{ checksum ".python.installed" }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
 
   - &save_deps_cache
     name: Saving Python dependency cache
-    key: v3-requirements-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
+    key: v3-requirements-{{ checksum ".python.installed" }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements-dev.txt" }}
     paths:
       - .venv
 
@@ -239,6 +239,9 @@ jobs:
           POSTGRES_PASSWORD: postgres
     steps:
       - checkout
+      - run:
+          name: Python version
+          command: python3 --version > .python.installed
       - restore_cache: *restore_deps_cache
       - run:
           name: Run kinto_remote_settings plugin unit tests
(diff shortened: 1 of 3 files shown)


grahamalama (commented):
Are we still exploring removing `uwsgi` as a dependency in `requirements.txt` and just installing it directly in the server Dockerfile? Or is that for a separate PR?

leplatrem:
> Are we still exploring removing uwsgi as a dependency in requirements.txt and just installing it directly in the server Dockerfile? Or is that for a separate PR?

I gave up with this idea because: if the version of Python on which the uwsgi Debian package was built is different that the virtualenv, then the packages are ignored.

grahamalama (approved):
> if the version of Python is upgraded

Did we see this happen in `cimg/python:3.10`? It looks like the [last release](https://hub.docker.com/layers/cimg/python/3.10/images/sha256-2916aa1ec88f5ca265b8810777cfd40189eef1278aa45b8a6d61e64da4a68686?context=explore) was 3 months ago, which seems a little earlier than when we started seeing failures.

Regardless, busting the virtualenv cache on the python version does seem like a good idea 👍🏻.

Not to address in this PR, but thinking out loud: Maybe it would be worth running `lint_format` and `unit_test` on `python:3.10.5-bullseye` (with the same SHA) so we have parity with the Python version we use to build the server.

grahamalama:
> Would that mean executing these from a container instead of from the Makefile ?

I think we'd still use the Makefile, just swap out the docker image we use to install the venv on top of and run the command. Something like this:

```diff
diff --git a/.circleci/config.yml b/.circleci/config.yml
index 99736e8..eb4749a 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,7 +231,7 @@ jobs:
 
   unit_test:
     docker:
-      - image: cimg/python:3.10
+      - image: python:3.10.5-bullseye
       - image: cimg/postgres:12.8
         environment:
           POSTGRES_DB: testdb
@@ -250,7 +250,7 @@ jobs:
 
   lint_format:
     docker:
-      - image: cimg/python:3.10
+      - image: python:3.10.5-bullseye
     steps:
       - checkout
       - run:
@@ -264,7 +264,7 @@ jobs:
 
   docs_test:
     docker:
-      - image: cimg/python:3.10
+      - image: python:3.10.5-bullseye
     steps:
       - checkout
       - restore_cache:
```

Though the `cimg/python` container has a lot of extras installed in it compared to the one published by Python, so I don't know if it would be a simple drop in replacement

leplatrem:
Oh, I see now. Do you want to try your changes in another PR?
I'll merge this one

grahamalama:
> Oh, I see now. Do you want to try your changes in another PR? I'll merge this one

Yep, I'll try it out

PR #257, opened 2022-08-01
Fix ordered list of scheduled jobs entry in support page

leplatrem:
Before
Screenshot 2022-08-01 at 19 13 50

After
Screenshot 2022-08-01 at 19 10 11


grahamalama (approved):
Nit, but would it make more sense to say "Ops" or "SRE" instead of "OPs"? Maybe it's just be, but when I first saw "OPs" I thought "`O`riginal `P`oste`s`"

PR #284, opened 2022-09-14
Check the Autograph cert validity in the heartbeat endpoint

leplatrem:
Since Autograph is not reachable from Telescope, I migrated the [certificate expiration check](https://github.com/mozilla-services/telescope/blob/main/checks/core/certificate_expiration.py) to the signer heartbeat


kinto-remote-settings/tests/signer/test_plugin_setup.py
```diff
@@ -79,26 +80,53 @@ def test_capability_is_exposed(self):
...
 
     def test_heartbeat_fails_if_missing_attributes(self):
         invalid = self.signature.copy()
         invalid.pop("signature")
-        self.mock.post.return_value.json.return_value = [invalid]
+        self.request_mock.post.return_value.json.return_value = [invalid]
+        resp = self.app.get("/__heartbeat__", status=503)
+        assert resp.json["signer"] is False
+
+    def test_heartbeat_fails_if_certificate_expires_soon(self):
```
grahamalama, line 124: Should we add test cases for all of the places the relative minimum can be in regards to the bounds? e.g. 
- no error when when `relative_minimum` > `max_remaining_days`
- error when `min_remaining_days` < `relative_minimum` <= `max_remaining_days`
- error when `min_remaining_days` >= `relative_minimum`

grahamalama (approved):
LGTM, just one pedantic comment about testing.

Also, we may want to consider using the word `clamp`/`clamped` instead of `bound`/`bounded`, since I think `clamp` is a commonly used name for what we're doing in other languages (CSS, lodash, and C++ have "clamp" functions)

PR #287, opened 2022-09-21
Bump 29.1.1

leplatrem:


.circleci/config.yml
```diff
@@ -152,6 +152,14 @@ jobs:
       BUILDKIT_PROGRESS: plain
     steps:
       - checkout
+      - run:
+          name: Check VERSION file
+          command: |
+            if [ "$CIRCLE_TAG" != $(cat VERSION) ];
```
smarnach, line 158: The `build_containers` job is run on all builds, but this check should only happen on tag builds, so you need to modify the condition here. I also recommend double quotes around `$(cat VERSION)`.

smarnach (approved):

PR #295, opened 2022-09-30
Add `Cache-Control` headers on all `/changeset` endpoints

leplatrem:
With this pull-request, all `/changeset` endpoints will have their `Cache-Control` response header value set based on the settings, in this order of precedence:

* `kinto.{bid}.{cid}.record_cache_maximum_expires_seconds` if `?_expected` is in querystring, `kinto.{bid}.{cid}.record_cache_expires_seconds` otherwise
* `kinto.{bid}.record_cache_expires_seconds` 
* `kinto.record_cache_expires_seconds` 

This mimics the current behavior of `/records`, as implemented here:
https://github.com/Kinto/kinto/blob/f996279cdcdf9e2eedf268b71a1fbb7ebdac67ce/kinto/views/records.py#L106-L133

> Note for Sven:  The `max-age=86400` is set on redirects, when `?_since=` contains a very old value (default value of `kinto.changes.since_max_age_redirect_ttl_seconds` setting).

* [ ] *TODO* Figure out where/how the `no-store` is coming from

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,15 @@ CHANGELOG
 This document describes changes between each past release as well as
 the version control of each dependency.
 
+29.1.2 (unreleased)
+===================
+
+**Bug Fixes**
+
+- Add a default ``Cache-Control`` header value on all ``/changeset`` endpoints using the
+  ``kinto.record_cache_expires_seconds`` setting value.
+
+
 29.1.1 (2022-09-21)
 ===================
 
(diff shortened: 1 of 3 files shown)


smarnach (approved):

smarnach:
I noticed that we also don't return a Cache-Control header for `/v1/`, which makes up about 15 percent of our requests. We definitely need to cache that endpoint as well, at least for some amount of time.

PR #306, opened 2022-10-17
Upgrade to kinto 15

leplatrem:
⚠️ **Breaking changes in configuration (`.ini`)**

- Switch from `raven` to `sentry-sdk`


leplatrem:
For a reason that I don't explain, some hashes were missing. For example, cornice-swagger did not have the hash for wheel sha256
Screenshot 2022-10-17 at 18 43 02

Also, several libraries are not up-to-date. Dependabot is skipping them?

grahamalama (approved):

grahamalama:
> Dependabot is skipping them?

Do hashes possibly change when we bump Python versions, like in #302?

leplatrem:
> Do hashes possibly change when we bump Python versions, like in https://github.com/mozilla/remote-settings/pull/302?

Hashes don't change by python version, since in the end it's just a hash of a downloaded package. 

Depending on some parameters (platform, cpu arch, ...) pip could pick either the `.tar.gz` or the `.whl` file. 
These rules may change from a pip version to another. And I just realize that a new pip version was released 2 days ago:
https://pip.pypa.io/en/stable/news/#v22-3
(I don't see anything in the changelog that is obviously related though)

PR #308, opened 2022-10-17
Return 200 instead of 304 to old clients reaching out monitor/changes

leplatrem:


smarnach:
I realized a problem with the approach of simply returning a 200: Cloud CDN cannot include the If-None-Match header in the cache key. This means it will simply cache the first 200 it sees for every `/records` endpoint, and return that for subseqauent requests. It might cache an empty list of changes and return that when it should actually return a more complete list of changes, or vice versa. This problem actually already exists, even without this PR. I'll try to figure out a way to fix this in Cloud CDN, but I'm not sure it's possible. We may need to introduce another caching layer in GCP.

leplatrem:
> This means it will simply cache the first 200 it sees for every /records endpoint,

oh, I see. But this will concern only clients reaching out without `?_since` queryparam. 
We can also change the strategy here. Instead of returning an empty list, we can return the whole list of changes then.

On the client side  (< Firefox 88, April 2021), this means they will iterate this list, and compare their local timestamp for every collection (60 extra reads to indexeddb...). Since all collections will be up-to-date, no extra sync will be initiated.

smarnach:
Turns out I misunderstood how exactly the If-None-Match header is interpreted by the client, so my last concern is moot. Let's go ahead and get this deployed. :)

grahamalama (approved):

smarnach (approved):

PR #317, opened 2022-10-26
Remove useless greenlet dependency

leplatrem:
This is a dependency that was previously part of SQLAchemy dependencies, but only used in the context of asyncio.


leplatrem:
`pip-compile` removes `greenlet` on my machine, because the platform does not match these conditions:

https://github.com/sqlalchemy/sqlalchemy/blob/8aa07c8e62a0483e0c882ffbf97453d5e4614efe/setup.cfg#L45

And this is the rationale https://github.com/sqlalchemy/sqlalchemy/issues/6136#issuecomment-896960050

PR #351, opened 2023-01-11
Bump v31.0.1

leplatrem:
* [x] Confirm that signatures are valid in DEV using devtools


grahamalama:
Did we
> Confirm that signatures are valid in DEV using devtools
?

leplatrem:
There aren't signatures anymore in DEV.

The Remote Settings Devtools UI shows a warning because the timestsamps from monitor/changes don't match. This is not great, although it's only visual and easy to fix.

grahamalama (approved):

leplatrem:
> This is not great, although it's only visual and easy to fix.

Actually, not so easy. The value shown as `server timestamp` in the DevTools comes from a call that is made by the client itself using `await RemoteSettings.inspect()`
https://searchfox.org/mozilla-central/rev/fb9a504ca73529fa550efe488db2a012a4bf5169/services/settings/remote-settings.js#509

We can either:

Change the client code of `inspect()` (must ride trains) etc.:
```diff
- const { bucket, collection, last_modified: serverTimestamp } = change;
+ const { bucket, collection, last_publication: serverTimestamp } = change;
```

Or change the DevTools code, to fetch the list of changes from the server instead of relying on `inspect()`. 

Or change the DevTools code, to just remove this warning.

I would vote to postpone the deployment of this v31 version in STAGE & PROD, until we have a new version of the DevTools released with a fix. Otherwise it could be relatively confusing for QA to see synchronization warnings when everything is actually up-to-date and successful.

leplatrem:
I will reopen a similar release PR after https://github.com/mozilla/remote-settings/pull/354 is merged

PR #354, opened 2023-01-12
Rollback changeset changes of #328 and #350

leplatrem:
# In #328 

The timestamp in the changeset view was from the collection metadata. But the payload sent to Autograph was using the records timestamp. Clients could never verify signatures, since they were building the payload using what was provided in the changeset.

# In #350

The timestamp in the changeset was changed to be the records timestamp. But since the timestamp of the monitor/changes is the collection metadata timestamp, the clients would never be up-to-date after sync. Therefore, on each sync, the clients would pull empty changesets and reverify local signature of all collections. On each sync. 

# Options...

The implementation of the chosen *Option 2* from the ADR cannot be implemented. 

To fix #328 and #330, we would have to send a payload to Autograph with the timestamp that is going to be the one we obtain *after* the obtained signature is stored in the collection metadata. Time travel not yet invented.

Since this took unexpected proportions, I propose that we rollback this work for now, and think of possible options later.

We can release a 31.0.1 that will yank 31.0.0. Its diff with 30.1.1 will be empty.


CHANGELOG.rst
```diff
@@ -4,21 +4,23 @@ CHANGELOG
...
+31.0.0 (yanked)
+===============
+
+**Yanked**
 
 **API changes**
 
 - The timestamps of the monitor/changes entries now reflect the timestamp of the collection metadata (eg. last signature) instead of the records timestamp (eg. last publication).
-  This change will **impact both servers and clients significantly**. See the `related ADR`<https://github.com/mozilla/remote-settings/blob/main/docs/adr/adr_002_cache_bust.md>`_ for more details.
+  This change will **impact both servers and clients significantly**. See the [related ADR](https://github.com/mozilla/remote-settings/blob/main/docs/adr/adr_002_cache_bust.md) for more details.
```
grahamalama, line 23: We actually want to keep this change, right?

grahamalama (approved):

leplatrem:
I removed the lines completely, to reduce any source of confusion, what do you think?

PR #362, opened 2023-02-03
Remote Rust toolchains, make other Docker / docker compose changes

grahamalama:
In this PR, we pulled in the latest `cannonicaljson-rs` wheels in `requirements.txt`.
Because of this change, we could remove the Rust toolchains from our Dockerfiles. While I was here, I made some other minor Dockerfile changes, as described in bf8ffba6f0410892625aa59e43cf5b638e280745.


bin/update_and_install_system_packages.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
leplatrem, line 1: Could you add a quick line here that this script takes the list of parameters as input?

leplatrem (approved):

leplatrem (approved):
This latest form makes sense 💯 

For the record, why was it important to exclude the `tests/` directory from the app container? Have you considered adding it in order to share the same dockerignore?

grahamalama:
> why was it important to exclude the tests/ directory from the app container?

I wouldn't say it was _highly_ important, just a matter of taste. Don't include what's not needed in either container, ya know?

Issue #368, opened 2023-02-07
Latest tag won't start


jaredlockhart:
The latest tag pushed to dockerhub here:

https://hub.docker.com/layers/mozilla/remote-settings/latest/images/sha256-96f0e2d2ef6afa2748b1b987f2aa910b724b0b1c4e9e10716890d23d391edc5d?context=explore

Won't start in my local or CI:

![image](https://user-images.githubusercontent.com/119884/217306936-fef23adb-d7e6-452c-be93-48739542e74e.png)

I don't see any update instructions in the tags:

https://github.com/mozilla/remote-settings/tags

grahamalama:
Looks like a file permissions issue in the container.

<details><summary>Traceback</summary>
<p>

```
Traceback (most recent call last):
  File "app.wsgi", line 30, in <module>
    application = main(config.items('DEFAULT'), **dict(config.items('app:main')))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/kinto/__init__.py", line 51, in main
    kinto.core.initialize(config, version=__version__, default_settings=DEFAULT_SETTINGS)
  File "/opt/venv/lib/python3.11/site-packages/kinto/core/initialization.py", line 618, in initialize
    config.include("kinto.core", route_prefix=api_version)
  File "/opt/venv/lib/python3.11/site-packages/pyramid/config/__init__.py", line 666, in include
    c(configurator)
  File "/opt/venv/lib/python3.11/site-packages/kinto/core/__init__.py", line 218, in includeme
    config.include(app)
  File "/opt/venv/lib/python3.11/site-packages/pyramid/config/__init__.py", line 666, in include
    c(configurator)
  File "/opt/venv/lib/python3.11/site-packages/kinto_emailer/__init__.py", line 192, in includeme
    config.include('pyramid_mailer' + ('.debug' if debug else ''))
  File "/opt/venv/lib/python3.11/site-packages/pyramid/config/__init__.py", line 666, in include
    c(configurator)
  File "/opt/venv/lib/python3.11/site-packages/pyramid_mailer/debug.py", line 8, in includeme
    mailer = DebugMailer(path)
             ^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.11/site-packages/pyramid_mailer/mailer.py", line 40, in __init__
[...]

PR #369, opened 2023-02-07
Make workdir (`/app`) home dir for user `app`

grahamalama:
Fixes: #368

### DIFF
--- a/RemoteSettings.Dockerfile
+++ b/RemoteSettings.Dockerfile
@@ -24,12 +24,13 @@ RUN /opt/update_and_install_system_packages.sh \
     # Needed for psycopg2
     libpq-dev
 
-RUN groupadd --gid 10001 app && \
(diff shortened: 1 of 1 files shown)


grahamalama:
@leplatrem should we add something to our integration test setup on top of this PR that would catch a problem like this in the future?

leplatrem:
> @leplatrem should we add something to our integration test setup on top of this PR that would catch a problem like this in the future?

I wonder why the browser tests did not catch that?

The integration tests only check for the heartbeat. We could add file permissions checks if necessary

leplatrem (approved):

grahamalama:
@leplatrem 
> I wonder why the browser tests did not catch that?

I think it's because we use Docker Compose to run the tests, which mounts a volume that doesn't experience such permissions issues.

PR #374, opened 2023-02-10
Enable code coverage report in tests

leplatrem:


kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
```diff
@@ -630,7 +630,7 @@ def cleanup_preview_destination(event, resources):
             continue
 
         for k in ("preview", "destination"):
-            if k not in resource:  # pragma: nocover
+            if k not in resource:  # pragma: no cover
```
grahamalama, line 633: I noticed that we used `nocover` in a few other places. If this is the correct way to mark a line as no cover, can we do a global find + replace?

Makefile
```diff
@@ -43,7 +44,8 @@ lint: $(INSTALL_STAMP)
 	$(VENV)/bin/flake8 kinto-remote-settings tests
 
 test: $(INSTALL_STAMP)
-	PYTHONPATH=. $(VENV)/bin/pytest kinto-remote-settings
+	PYTHONPATH=. $(VENV)/bin/coverage run --rcfile pyproject.toml -m pytest kinto-remote-settings
```
grahamalama, line 47: Do we need `--rcfile pyproject.toml`? Does coverage not find it automatically?

kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py
grahamalama: Also, should we not cover this line, or should we add a test case that covers this branch?

leplatrem: It is relatively acceptable to ignore this branch. 

To cover it, we should add a test that cleanups a resource which hasn't any preview (`source -> destination`)

grahamalama (approved):

PR #378, opened 2023-02-15
Use Docker-managed data volumes, profiles for managing services, other Docker tweaks

grahamalama:
This PR:
- specifies [profiles](https://docs.docker.com/compose/profiles/) for Docker compose so that, among other things, running `make start` (`docker compose up`) doesn't also spin up Selenium and the tests container

- Modifies our compose / Dockerfile setup so that we can rely on just Docker volumes to manage data for local development and integration tests
There isn't a super clean way to control permissions for named volumes, and this does add a little complexity to our Docker setup overall. But, I think it is nicer to have docker-compose manage volumes and to not have test data end up in the root directory.

- moves building the main `kinto-remote-settings` plugin to the `compile` stage
This means we don't have do repeat this step every time we change something in the project directory, since we'd first `COPY . .` before bulding the plugin

- tweaks the way we specify `ENTRYPOINT` and `CMD` instructions

### DIFF
--- a/IntegrationTests.Dockerfile
+++ b/IntegrationTests.Dockerfile
@@ -14,5 +14,5 @@ RUN pip install --no-cache-dir -r /opt/requirements.txt
 
 WORKDIR /app
 COPY tests/ pyproject.toml ./
-
+# ./tests/run.sh, not ./bin/run.sh
 ENTRYPOINT ["/app/run.sh"]
--- a/IntegrationTests.Dockerfile
+++ b/IntegrationTests.Dockerfile
@@ -15,4 +15,4 @@ RUN pip install --no-cache-dir -r /opt/requirements.txt
 WORKDIR /app
 COPY tests/ pyproject.toml ./
 
-ENTRYPOINT ["/bin/bash", "/app/run.sh"]
+ENTRYPOINT ["/app/run.sh"]
--- a/IntegrationTests.Dockerfile
+++ b/IntegrationTests.Dockerfile
@@ -13,6 +13,6 @@ COPY tests/requirements.txt /opt
 RUN pip install --no-cache-dir -r /opt/requirements.txt
 
 WORKDIR /app
-COPY tests .
+COPY tests/ pyproject.toml ./
 
 ENTRYPOINT ["/bin/bash", "/app/run.sh"]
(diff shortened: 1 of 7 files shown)


RemoteSettings.Dockerfile
```diff
@@ -34,13 +35,16 @@ RUN chown 10001:10001 /app && \
...
 USER app
+ENTRYPOINT ["./bin/run.sh"]
 # Run uwsgi by default
-ENTRYPOINT ["/bin/bash", "/app/bin/run.sh"]
-CMD ["sh", "-c", "uwsgi --http :${PORT} --ini ${KINTO_INI}"]
+CMD ["start"]
+
+FROM production as local
+# create directories for volume mounts used in integration tests / local development
+RUN mkdir -m 777 /app/mail && mkdir -m 777 /tmp/attachments
```
leplatrem, line 50:
```
mkdir -m 777 /app/mail && mkdir -m 777 /tmp/attachments:
#0 0.054 mkdir: cannot create directory ‘/app/mail’: File exists
```

Would maybe require a `-p` option

IntegrationTests.Dockerfile
```diff
@@ -13,6 +13,6 @@ COPY tests/requirements.txt /opt
 RUN pip install --no-cache-dir -r /opt/requirements.txt
 
 WORKDIR /app
-COPY tests .
+COPY tests/ pyproject.toml ./
 
-ENTRYPOINT ["/bin/bash", "/app/run.sh"]
+ENTRYPOINT ["/app/run.sh"]
```
leplatrem, line 18: Could you add a comment that this is `tests/run.sh` in the repo? (and not `bin/run.sh`)

leplatrem (approved):
Neat!

Thanks for taking the time to tweak this 💯 

I could test it locally, code changes were taken into account without rebuilding the world 👌

RemoteSettings.Dockerfile
grahamalama: [54d2845](https://github.com/mozilla/remote-settings/pull/378/commits/54d284539449e7e583f564577e30b2ac35cf507e)

IntegrationTests.Dockerfile
grahamalama: [3f703a9](https://github.com/mozilla/remote-settings/pull/378/commits/3f703a9d11e237205cc6d4a5c31426289e70ae49)

PR #380, opened 2023-02-16
Prevent browser tests to be skipped

leplatrem:
Before 

Screenshot 2023-02-16 at 15 42 40

After

Screenshot 2023-02-16 at 17 27 48


grahamalama:
It looks like this test skipped initially because `sensitiveurl` matched `.*`, but in the second test run, it was properly set to `settings-writer.prod.mozaws.net/v1/admin`.

Do we want to actually make sure this test doesn't run against the prod environment? If so, we may want to be sure `sensitiveurl` is properly set rather than saying this is safe to run in any case.

grahamalama (approved):
Running `make browser-test` did successfully set `sensitiveurl`
image

Still not sure why yours was set to `.*` in your first screenshot, but this seems like it worked?

PR #394, opened 2023-03-22
Use poetry

grahamalama:
Migrate from `pip-tools` to Poetry for dependency management

- [x] format toml
- [x] incorporate into CircleCI ~_or_ move linting / tests to GHA~
  - it might be good to move everything onto GHA in a separate PR. Or not.
- [x] Fine-tune poetry installation, commands, dependency groups
- [x] Update README
- [x] Update CHANGELOG


.readthedocs.yaml
grahamalama, line 1: Taken from: https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry

.circleci/config.yml
```diff
@@ -266,17 +260,17 @@ jobs:
...
     docker:
-      - image: python:3.10.5-bullseye@sha256:dac61c6d3e7ac6deb2926dd96d38090dcba0cb1cf9196ccc5740f25ebe449f50
+      - image: cimg/python:3.11.2
     steps:
       - checkout
       - restore_cache:
-          name: Restoring Python docs dependency cache
-          key: v1-requirements-{{ checksum "docs/requirements.txt" }}
+          name: Restoring docs dependency cache
+          key: v2-docs-requirements-{{ checksum "poetry.lock" }}
```
grahamalama, line 268: This will unfortunately bust the cache more than just the checksum of `docs/requirements.txt`, but hopefully this process will still go quickly since we only have 2 dependencies in the docs group.

.readthedocs.yaml
grahamalama, line 1: Is there a way I can test that this actually works before merging?

.circleci/config.yml
```diff
@@ -233,17 +233,14 @@ jobs:
 
   unit_test:
     docker:
-      - image: python:3.10.5-bullseye@sha256:dac61c6d3e7ac6deb2926dd96d38090dcba0cb1cf9196ccc5740f25ebe449f50
+      - image: cimg/python:3.11.2
```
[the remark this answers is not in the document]
grahamalama: There is, but I thought it'd be good to make sure we use the same patch version that we use in our Dockerfile. Or is that being too careful?

.readthedocs.yaml
grahamalama: Nice -- looks like that worked! https://readthedocs.org/projects/remote-settings/builds/20015230/

.circleci/config.yml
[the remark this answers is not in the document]
leplatrem: Being careful is good :) I was thinking that we could have a test that verifies that both versions are in sync. I doubt that dependabot will upgrade this line. Will it?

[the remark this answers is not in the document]
grahamalama: Yeah, I don't think it will update this line. 

We're actually in an interesting situation now where there's a `python:3.11.3` published, but there isn't a `cimg/python:3.11.3` yet

RemoteSettings.Dockerfile
```diff
@@ -1,24 +1,39 @@
...
 WORKDIR /opt
-COPY requirements.txt .
-RUN pip install --no-cache-dir -r requirements.txt && \
+COPY ./poetry.lock ./pyproject.toml ./
+RUN $POETRY_HOME/bin/poetry install --only main --no-root && \
     uwsgi --build-plugin https://github.com/Datadog/uwsgi-dogstatsd
+
+# though we have kinto-remote-settings specified as a dependency in
+# pyproject.toml, we have it configured to install in editable mode for local
+# development. For building the container, we only install the "main"
```
leplatrem, line 23: > in editable mode for local development

because we use `--no-root` in `make install`? and `develop = true` in pyproject.toml?

leplatrem (approved):
Thanks for your efforts on this one 👏

RemoteSettings.Dockerfile
grahamalama: I think `develop = true` is more important than `--no-root` in this context

.circleci/config.yml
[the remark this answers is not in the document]
grahamalama: https://github.com/mozilla/remote-settings/issues/406

PR #401, opened 2023-03-29
Fix #400: rely on `/changeset` endpoint for signature integration tests

### LINKED ISSUE #400: Rely on `/changeset` endpoints in integration tests
This would match our clients behavior

leplatrem:
Fix #400

### DIFF
--- a/tests/integration/plugins/test_signer.py
+++ b/tests/integration/plugins/test_signer.py
@@ -97,8 +97,9 @@ async def test_signer_plugin_full_workflow(
         # Re-sign and verify.
         await editor_client.patch_collection(data={"status": "to-review"})
         await reviewer_client.patch_collection(data={"status": "to-sign"})
-        timestamp = await dest_client.get_records_timestamp()
-        signature = (await dest_client.get_collection())["data"]["signature"]
+        changeset = await dest_client.fetch_changeset()
+        timestamp = changeset["timestamp"]
+        signature = changeset["metadata"]["signature"]
         await verify_signature([], timestamp, signature)
 
     # 1. upload data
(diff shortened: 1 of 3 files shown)


tests/integration/plugins/test_signer.py
```diff
@@ -39,6 +39,15 @@ async def verify_signature(records, timestamp, signature):
         await verifier.verify(serialized, signature["signature"], x5u)
 
 
+def fetch_changeset(client, bucket=None, collection=None, **kwargs):
+    bucket = bucket or client.bucket_name
+    collection = collection or client.collection_name
```
grahamalama, line 44: Is `client` guaranteed to have `bucket_name` and `collection_name` as a fallback?

tests/integration/plugins/test_signer.py
```diff
@@ -39,6 +39,15 @@ async def verify_signature(records, timestamp, signature):
...
+    endpoint = f"/buckets/{bucket}/collections/{collection}/changeset"
+    kwargs.setdefault("_expected", random.randint(999999000000, 999999999999))
```
grahamalama, line 46: What does this `_expected` kwarg do? Only curious for my understanding.

tests/integration/plugins/test_signer.py
```diff
@@ -39,6 +39,15 @@ async def verify_signature(records, timestamp, signature):
         await verifier.verify(serialized, signature["signature"], x5u)
 
 
+def fetch_changeset(client, bucket=None, collection=None, **kwargs):
```
grahamalama, line 42: Does is matter that this function isn't `async`?

grahamalama (approved):

tests/integration/plugins/test_signer.py
leplatrem: We could just remove the `bucket` and `collection` arguments, since we pass a client that has those attributes set at instantiation

tests/integration/plugins/test_signer.py
leplatrem: we use it for cache bust. On client, it contains the highest timestamp of the monitor/changes list, and is received through the push notification.
here we could pass `?_expected=0` it would be fine. Using a random one just makes sure we're not going to be hitting cache during tests

tests/integration/plugins/test_signer.py
leplatrem: Well. Alternatives would be:
- inherit and add a method to the `AsyncClient`
- add a `await loop.run_in_executor(None, lambda: fetch_changeset(...))`

What do you think?

grahamalama: I think ultimately we should add a method to the `AsyncClient`, but we can do your second suggestion to get this merged as long as we get a follow-on issue.

PR #407, opened 2023-04-06
Dependabot tweaks

grahamalama:
- Remove checks for dependencies in `tests` directory
- Remove ignored updates, since all of the dependencies are greater than those that are ignored


leplatrem (approved):
👍

PR #411, opened 2023-04-12
Fixes #410: Reorder float error message so path is at the end

### LINKED ISSUE #410: Improve error message when request contains float values
When the HTTP PUT/POST body contains a `float` we reject it with an [HTTP 400 Bad request error](https://github.com/mozilla/remote-settings/blob/9bdc3842511b8d5bb4169bb19aad306a2ff784ef/kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py#L402).  

Since `scan()` is recursive, and if it is in a deeply nested data structure,  the exception can be useless for troubleshooting.  For example, Sentry truncates long error messages: 

image

Maybe a good fix is to reorder the feedback from [this line](https://github.com/mozilla/remote-settings/blob/9bdc3842511b8d5bb4169bb19aad306a2ff784ef/kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py#L382): 

* from `f"'{path}' field contains float value (tip: use integer or string)"` 
* to `f"field contains float value (tip: use integer or string), '{path}'"`

mostlygeek:
The `path` where a float is detected can be quite long.  Some tools like Sentry can truncate the error message so the useful error message is removed.

This makes troubleshooting more challenging and less obvious.  Moving the path to the end makes it clearer what the error is and where to start looking for the problem.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,10 @@ the version control of each dependency.
 31.0.2.dev (unreleased)
 ===================
 
+**Bug Fixes**
+
+- Fixes #410: Reorder float error message so path is at the end
+
 **Internal Changes**
 
 - Improve Docker volume management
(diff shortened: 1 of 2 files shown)


bsieber-mozilla (approved):
LGTM

for some reason the CI job didn't run with your branch; so I forked it--and it re-ran with success
Screenshot 2023-04-12 at 3 01 17 PM

mostlygeek:
Thanks.

PR #423, opened 2023-06-05
Add Dependabot automation workflow

grahamalama:


grahamalama:
image

This setting is currently enabled, which prevents us from using this Dependabot automation.

To get around this we could:
- uncheck that box
- change the codeowners file from 
```
*        @mozilla/remote-settings-reviewers
```
to something more granular that doesn't cover files Dependabot would touch
- something else?

leplatrem (approved):

grahamalama:
I'll merge this and follow up with how we can address the Codeowners blocker

Issue #475, opened 2023-10-16
Migrate CI/CD pipeline from CircleCI to Github Actions


grahamalama:
```[tasklist]
### Tasks
- [ ] https://github.com/mozilla/remote-settings/pull/477
- [x] Update brach protection to reference GHA CI checks (instead of CIrcle)
- [ ] https://github.com/mozilla/remote-settings/pull/478
- [ ] https://github.com/mozilla/remote-settings/pull/482
- [ ] https://github.com/mozilla/remote-settings/issues/489
```

leplatrem:
This was done 💯

grahamalama:
@leplatrem I'm going to reopen this since I want to consider it done when we publish directly to the GAR (I have it as an open issue in the task list above)

PR #477, opened 2023-10-18
Migrate CI checks to GHA

grahamalama:
Ref: #475


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - uses: actions/cache@v3
+        id: cache-poetry
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MIN: "15"
+        with:
+          path: |
+            ~/.cache/pypoetry/virtualenvs
+            ~/.cache/pypoetry/cache
+            ~/.cache/pypoetry/artifacts
+          key: poetry-dev-${{ hashFiles('poetry.lock') }}
```
grahamalama, line 21: We need to use this caching strategy (rather than setting `cache: poetry` in `setup-python`) because our use of dependency groups. We want separate caches for each dependency group.

https://github.com/actions/setup-python/issues/505

docker-compose.yml
```diff
@@ -50,6 +50,7 @@ services:
     build:
       dockerfile: RemoteSettings.Dockerfile
       target: local
+      context: .
```
grahamalama, line 53: For some reason, GHA threw an error without this option set in the Dockerfile (same for the tests container)

.github/workflows/ci.yaml
grahamalama: (Going to include this as a comment)

PR #482, opened 2023-10-25
Add caching for web / test container builds in CI

grahamalama:
This PR adds caching for container builds for our integration and browser tests in CI.

In a recent PR on our `main` branch, browser and integration tests took roughly 5m 15s to run ([src](https://github.com/mozilla/remote-settings/actions/runs/6642086964))

On the first, uncached run of this workflow (in a fork), [we can see](https://github.com/grahamalama/remote-settings/actions/runs/6644408410) that about 2m 47s are dedicated to building our containers (minus some time to persist layers to the cache).

image

But [after the cache is populated](https://github.com/grahamalama/remote-settings/actions/runs/6644513916), that build time drops down to ~47s

image

While this is a nice improvement, one issue with the current approach is that the size of the cache is quite large at around ~770mb

image

Since Github Actions only provides 10GB of cache storage per repo, we'll start running out of cache room rather quickly. Github will start evicting our oldest caches to make room for new ones, which could cause unexpected slow builds. ([docs](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy))

Maybe there's a better caching strategy than the one that's currently in this PR that balances cache size and usefulness. Happy to hear ideas.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -103,7 +103,8 @@ jobs:
       - run: mkdir ${DOCKER_CACHE}
         
       - name: Compute cache key
-        run: echo "CACHE_KEY=$(git ls-tree --full-tree -r --name-only HEAD | grep -v -f RemoteSettings.dockerignore | git hash-object --stdin)" >> $GITHUB_ENV
+        # Create hash of hashes of checked in files not in Dockerignore
+        run: echo "CACHE_KEY=$(git ls-tree --full-tree -r HEAD | grep -v -f RemoteSettings.dockerignore | awk '{print $3}' | git hash-object --stdin)" >> $GITHUB_ENV
 
       - uses: actions/cache@v3
         with:
@@ -146,7 +147,8 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Compute cache key
-        run: echo "CACHE_KEY=$(git ls-tree --full-tree -r --name-only HEAD | grep -v -f RemoteSettings.dockerignore | git hash-object --stdin)" >> $GITHUB_ENV
+        # Create hash of hashes of checked in files not in Dockerignore
+        run: echo "CACHE_KEY=$(git ls-tree --full-tree -r HEAD | grep -v -f RemoteSettings.dockerignore | awk '{print $3}' | git hash-object --stdin)" >> $GITHUB_ENV
 
       - uses: actions/cache@v3
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -95,16 +95,90 @@ jobs:
         run: make test
   integration_test:
     runs-on: ubuntu-latest
+    env:
+      DOCKER_CACHE: /tmp/docker-cache
     steps:
       - uses: actions/checkout@v3
+
+      - run: |
+          mkdir -p ${DOCKER_CACHE}/web
+          mkdir -p ${DOCKER_CACHE}/tests
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -23,7 +23,7 @@ maintainer-clean: distclean ## Delete all non versioned files
 	rm -rf .pytest_cache
 	rm -rf tests/.pytest_cache
 	find . -name '*.orig' -delete
-	docker-compose down --remove-orphans --volumes --rmi all
+	docker compose down --remove-orphans --volumes --rmi all
```
grahamalama, line 26: [About all of these changes](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose). I know we've dealt with this in other repos by looking for both `docker compose` and `docker-compose` and putting that command into a variable, but maybe we should just force the switch here?

.github/workflows/ci.yaml
```diff
@@ -95,16 +95,90 @@ jobs:
...
+      DOCKER_CACHE: /tmp/docker-cache
     steps:
       - uses: actions/checkout@v3
+
+      - run: |
+          mkdir -p ${DOCKER_CACHE}/web
+          mkdir -p ${DOCKER_CACHE}/tests
+
+      - name: Compute cache key
+        run: echo "CACHE_KEY=$(git ls-tree --full-tree -r --name-only HEAD | grep -v -f RemoteSettings.dockerignore | git hash-object --stdin)" >> $GITHUB_ENV
```
grahamalama, line 108: Probably deserves a comment -- the intent is to create a new cache key only when a file that we use in our Docker context changes. So `hash(<files checked into git> - <files we ignore in the docker container>)`

Makefile
leplatrem: I'm good with using `docker compose` without a variable yes ;)

leplatrem:
> Maybe there's a better caching strategy than the one that's currently in this PR that balances cache size and usefulness. 

I think this improves the current situation enough to ship it. 

> Happy to hear ideas.

I can't remember in which project we used to dump the containers as binary files into cache with `docker save` and reload them with `docker load -i` in other steps. That would probably use less cache that caching every layer?

grahamalama:
> I can't remember in which project we used to dump the containers as binary files into cache with docker save and reload them with docker load -i in other steps. That would probably use less cache that caching every layer?

We could do basically the same thing if we omitted the `mode=max` flag from the `cache-to` args. But I think caching every layer is what we want, especially with our multistage builds. Both of our containers basically do, among other things:

```
FROM python:3.11.5-slim as build
... # create virtual env, install python dependencies

FROM python:3.11.5-slim
...

COPY --from=build $VIRTUAL_ENV $VIRTUAL_ENV
...
```

by caching only the final image's layers, we'd need to rebuild the `build` stage layers each time, and that's where most of the time-consuming work is done.

Another approach that I saw was to use [a container registry as a cache](https://docs.docker.com/build/cache/backends/registry/), rather than the Github Actions cache. This would give us more storage, but could come with its own considerations.

grahamalama:
Speaking of cache backends, I also tried out the [buildx gha](https://docs.docker.com/build/cache/backends/gha/) cache backend, and the results weren't as nice as using the local cache.

An initial push of a new branch would build everything from scratch (and presumably cache layers for that branch), but commits after the initial push would pull from the cache and be speedy.

leplatrem (approved):
:shipit:

PR #487, opened 2023-10-30
Replace dependabot automation action with groups feature

### LINKED ISSUE #486: Replace Dependabot automation with groups
We recently discovered that the action that we use to automatically approve and merge Dependabot PRs doesn't trigger our `publish` workflow when the PR is merged by the bot.

The goal of the automation is to reduce the amount of "touches" we need to merge PRs, so we're going to try using groups to see if it can achieve a similar outcome while also preserving that each commit to `main` triggers a workflow consistently.

grahamalama:
Closes #486

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,13 @@ updates:
     interval: weekly
     day: sunday
     timezone: UCT
+  groups:
+    production-dependency-patches:
+      dependency-type: "production"
+      update-types: ["patch"]
+    development-dependencies:
+      dependency-type: "development"
+      update-types: ["major", "minor", "patch"]
   open-pull-requests-limit: 99
   reviewers:
   - leplatrem
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -182,13 +182,3 @@ jobs:
 
       - name: Run browser tests
         run: make browser-test
-
-  review-dependabot-pr:
-    permissions:
-      contents: write
-      pull-requests: write
-    runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -182,13 +182,3 @@ jobs:
...
-
-  review-dependabot-pr:
-    permissions:
-      contents: write
-      pull-requests: write
-    runs-on: ubuntu-latest
-    if: ${{ github.actor == 'dependabot[bot]' }}
-    steps:
-      - name: Enable Dependabot automation
-        uses: mozilla/syseng-pod/actions/dependabot-automerge@main
```
leplatrem, line 194: 😢 
Since Dev dependencies are not "deployed", shall we keep this automerge for them?

.github/dependabot.yml
```diff
@@ -6,6 +6,13 @@ updates:
     interval: weekly
     day: sunday
     timezone: UCT
+  groups:
+    production-dependency-patches:
+      dependency-type: "production"
+      update-types: ["patch"]
```
leplatrem, line 12: Here we group all patches together, but minor and major will get independent requests.

If we don't increase the schedule to "weekly", then there are less likely to be grouped, no?

.github/workflows/ci.yaml
grahamalama: We set up a dependabot group for all development dependencies so while it's not automatic, it's still fewer clicks. 

I'm still overall wary about automatically merging code from a security perspective. Even running a compromised linting tool locally (for example) could pose a risk.

leplatrem (approved):

PR #492, opened 2023-11-03
Add docs about monitoring

leplatrem:
Let's document our queries somewhere. 

This is related to `telemetry-query-examples` of our [Quality Standards](https://mozilla.github.io/syseng-pod/quality-standards/)

PR #494, opened 2023-11-06
Fix authors string in pyproject.toml

leplatrem:
Pull-requests like #492 or #493 fail because the authors string isn't an email 🤷 

We don't have an official contact email, so at least let's be clear.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ asyncio_mode = "strict"
 name = "remote-settings"
 version = "0"
 description = "Easily manage evergreen settings data in Firefox"
-authors = ["@mozilla/remote-settings-reviewers"]
+authors = ["Contactless <postmaster@localhost>"]
 license = "MPL"
 readme = "README.rst"
 
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ down:  ## Shutwdown all containers
 
 install-docs: $(VENV)/bin/python $(DOC_STAMP)  ## Install documentation build dependencies
 $(DOC_STAMP): poetry.lock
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -42,7 +42,7 @@ asyncio_mode = "strict"
 name = "remote-settings"
 version = "0"
 description = "Easily manage evergreen settings data in Firefox"
-authors = ["@mozilla/remote-settings-reviewers"]
+authors = ["Contactless <postmaster@localhost>"]
```
grahamalama, line 45: This is a frankly `very annoying` feature of Poetry.

I think this is happening because even though `make install-docs` specifies we're only installing the docs dependencies, it still tries to install the root project.

We could also make this error go away with this change:

```diff
diff --git a/Makefile b/Makefile
index cb5751b..b51a823 100644
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ down:  ## Shutwdown all containers
 
 install-docs: $(VENV)/bin/python $(DOC_STAMP)  ## Install documentation build dependencies
 $(DOC_STAMP): poetry.lock
-	POETRY_VIRTUALENVS_IN_PROJECT=1 poetry install --only docs
+	POETRY_VIRTUALENVS_IN_PROJECT=1 poetry install --no-root --only docs
 	touch $(DOC_STAMP)
 
 docs: install-docs  ## Build documentation

```

grahamalama (approved):

pyproject.toml
leplatrem: I like your proposal, so I did both 😎

PR #498, opened 2023-11-13
Remove mentions of CircleCI

leplatrem:


docs/support.rst
```diff
@@ -150,7 +150,7 @@ See :ref:`multi-signoff tutorial <tutorial-multi-signoff-request-review>` for mo
 With regards to the repository:
 
 - MUST build a Docker container
-- MUST contain a CircleCI configuration that will publish to Dockerhub once credentials are setup by Ops
+- MUST contain a Github Action that will publish to Dockerhub once credentials are setup by Ops
```
grahamalama, line 153: TBD. This will probably be true, but the alternative would be that the publishing happens somewhere else (in the SRE deploy repo)

grahamalama (approved):

docs/support.rst
leplatrem: Well, this applies to ingestion jobs containers, I think it's fine to keep it simple enough and use Dockerhub

PR #500, opened 2023-11-14
Use `ruff` instead of `flake8`, `isort`, and `black`

leplatrem:


pyproject.toml
```diff
@@ -1,7 +1,36 @@
...
+# Exclude a variety of commonly ignored directories.
+exclude = [
+    ".git",
+    "__pycache__",
+    ".venv",
+    "node_modules",
+]
+
+[tool.ruff.lint]
+preview = true
```
grahamalama, line 11: Do we want this? Seems like it comes with less stability ([docs](https://docs.astral.sh/ruff/preview/))

pyproject.toml
```diff
@@ -1,7 +1,36 @@
-[tool.isort]
-profile = "black"
-lines_after_imports = 2
-src_paths = ["kinto-remote-settings", "tests"]
+[tool.ruff]
+# Exclude a variety of commonly ignored directories.
+exclude = [
+    ".git",
+    "__pycache__",
```
grahamalama, line 5: It seems that this is the only rule that's not in the default excluded directories. Maybe we can use [`extend-exclude`](https://docs.astral.sh/ruff/settings/#extend-exclude) instead?

pyproject.toml
```diff
@@ -1,7 +1,36 @@
...
+
+[tool.ruff.format]
+# Like Black, use double quotes for strings.
+quote-style = "double"
+# Like Black, indent with spaces, rather than tabs.
+indent-style = "space"
+# Like Black, respect magic trailing commas.
+skip-magic-trailing-comma = false
+# Like Black, automatically detect the appropriate line ending.
+line-ending = "auto"
```
grahamalama, line 33: This looks the same as what comes in the default configuration. Can we get away with not specifying it?

tests/integration/plugins/test_attachment.py
```diff
@@ -21,7 +21,7 @@ async def test_attachment_plugin_new_record(
     editor_client = make_client(editor_auth)
     with open("kinto-logo.svg", "rb") as attachment:
         assert requests.post(
-            f"{editor_client.session.server_url}{await editor_client.get_endpoint('record', id='logo')}/attachment",
+            f"{editor_client.session.server_url}{await editor_client.get_endpoint('record', id='logo')}/attachment",  # noqa: E501
```
grahamalama, line 24: Should we just ignore `E501` globally and let the formater take care of long lines when it can?

pyproject.toml
```diff
@@ -1,7 +1,36 @@
...
+
+[tool.ruff.lint]
+preview = true
+select = [
+    # pycodestyle
+    "E", "W",
+    # flake8
+    "F",
+    # isort
+    "I"
```
grahamalama, line 18: It seems that ruff comes with its own import sorter. Can we get away with not specifying `isort` linting rules and letting the formatter handle it all? 

I know we set `lines-after-imports = 2` below, but if we were fine without enabling that rule, it would be mean fewer lines of config overall.

tests/integration/plugins/test_attachment.py
grahamalama: Oh wait, just noticed https://github.com/mozilla/remote-settings/pull/500/commits/dccb698d3ad0bcac5dc7fc30d723b8dfe5422dd0. What motivated this commit?

pyproject.toml
leplatrem: nope 👍

pyproject.toml
leplatrem: My bad

tests/integration/plugins/test_attachment.py
leplatrem: We had `noqa` strings that ruff could not handle, I thought it was redundant and went for the route of keeping the lint rule, but yeah, let's handle this with format

pyproject.toml
leplatrem: > It seems that ruff comes with its own import sorter. 

Yes, it's enabled with `I` rules!

> if we were fine without enabling that rule, it would be mean fewer lines of config overall.

I'm +1 on keeping this config line, since it's the standard we use on our other repos

kinto-remote-settings/src/kinto_remote_settings/changes/views.py
```diff
@@ -40,8 +40,7 @@ def __init__(self, request):
     def timestamp(self):
         if not self._entries():
             return core_utils.msec_time()
-        max_value = max([e["last_modified"] for e in self._entries()])
-        return max_value
+        return max([e["last_modified"] for e in self._entries()])
```
grahamalama, line 43: The thing I like about this old version is that it makes it easier to throw a breakpoint in there before returning

```diff
  max_value = max([e["last_modified"] for e in self._entries()])
+ breakpoint()
  return max_value
```

is there a way to achieve something similar with this new linting rule?

tests/conftest.py
```diff
@@ -231,13 +230,13 @@ def verify_url(request: FixtureRequest, base_url: str):
         session.get(base_url, verify=False)
 
 
-@pytest.fixture
+@pytest.fixture()
```
grahamalama, line 233: What motivated these changes?

grahamalama (approved):
Nice. Let's rough it up everywhere. More ruff.


I was thinking we might want to squash this branch into 2 commits:
- one where we add ruff + the final config
- one where we make all of the changes

Then rebasing the branch on main (instead of squashing) so we can do something [like this](https://akrabat.com/ignoring-revisions-with-git-blame/) to ignore the mass formatting in git blame.

Not necessary by any means, but a nice-to-have.

kinto-remote-settings/src/kinto_remote_settings/changes/views.py
leplatrem: I hadn't thought of that. I could definitely revert 8264b5eb7450bf477dea46e15c8d57cb16273683

tests/conftest.py
leplatrem: # pytest-fixture-incorrect-parentheses-style (PT001)

Derived from the **flake8-pytest-style** linter.

Fix is always available.

## What it does
Checks for argument-free `@pytest.fixture()` decorators with or without
parentheses, depending on the `flake8-pytest-style.fixture-parentheses`
setting.

## Why is this bad?
If a `@pytest.fixture()` doesn't take any arguments, the parentheses are
optional.

Either removing those unnecessary parentheses _or_ requiring them for all
fixtures is fine, but it's best to be consistent.

## Example
```python
import pytest


@pytest.fixture
def my_fixture():
    ...
```

Use instead:
```python
import pytest


@pytest.fixture()
def my_fixture():
    ...
```

## Options
- `flake8-pytest-style.fixture-parentheses`

## References
- [`pytest` documentation: API Reference: Fixtures](https://docs.pytest.org/en/latest/reference/reference.html#fixtures-api)

leplatrem:
> Then rebasing the branch on main (instead of squashing) so we can do something [like this](https://akrabat.com/ignoring-revisions-with-git-blame/) to ignore the mass formatting in git blame.

Screenshot 2023-11-16 at 11 26 37

We can squash, ignoring rev would still be possible anyway

PR #512, opened 2023-12-05
Updating kinto to 16.2.3. Kinto-admin to 2.1.1

alexcottner:
Updating kinto to [v16.2.3](https://github.com/Kinto/kinto/releases/tag/16.2.3), which is updating kinto-admin to [v2.1.1](https://github.com/Kinto/kinto-admin/releases/tag/v2.1.1).

PR #520, opened 2024-01-09
Adding wait for browser tests

alexcottner:
Adding a 1 second wait to browser tests to wait for hooks to finish running.
Not 100% sure why this is an intermittent issue as I cannot reproduce the problem in my browser and reproducing locally was very difficult. Inspecting the DOM shows nothing is actually changing so likely hook state.

### DIFF
--- a/tests/browser_test.py
+++ b/tests/browser_test.py
@@ -1,4 +1,5 @@
 import pytest
+import time
 from kinto_http.patch_type import JSONPatch
 from selenium.common.exceptions import NoSuchElementException
 from selenium.webdriver.common.by import By
@@ -64,6 +65,8 @@ async def test_review_signoff(
         + f"#/buckets/{source_bucket}/collections/{source_collection}/simple-review"
     )
     selenium.refresh()
+    
+    time.sleep(1)  # give react a second for hooks
 
     try:
         approve_button: WebElement = selenium.find_element(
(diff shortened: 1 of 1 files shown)


tests/browser_test.py
```diff
@@ -65,6 +67,8 @@ async def test_review_signoff(
     )
     selenium.refresh()
 
+    time.sleep(1)  # give react a second for hooks
```
grahamalama, line 70: https://github.com/mozilla/remote-settings/pull/512/files/01698db34da090fbf923df2a87ee9a2f2466b601..5b1fda47fccfad660d6971331a120faf622f1c93#diff-03ae4d94c3127061bcc8c5d44da8fe1745f8fe0ff6d3018e3cd95f6c697b3df6

Everything old is new again 😄 .

Same comment as there -- can we use Selenium to [explicitly wait](https://www.selenium.dev/documentation/webdriver/waits/#explicit-waits) instead of `time`? Either for a duration or [expected condition](https://www.selenium.dev/documentation/webdriver/support_features/expected_conditions/).

grahamalama: Alright, that's annoying. Let's merge this to unblock ourselves, but can we make an issue to investigate this further?

grahamalama (approved):

tests/browser_test.py
alexcottner: created https://github.com/mozilla/remote-settings/issues/522 to track it. We'll have to do the same thing in kinto-admin probably.

PR #521, opened 2024-01-09
Add sequence diagrams to ADR 2

grahamalama:
I thought some sequence diagrams of the release process would be helpful for this ADR. Feel free to incorporate (or not) as you see fit


leplatrem:
Thanks!

PR #525, opened 2024-01-15
Pull Kinto Admin from sources

leplatrem:
* [x] verify default behaviour
* [x] check that admin runs at specified version (by downgrading it for example)
* [x] verify behaviour of env var vs. ini setting:  .ini setting has no effect if the env var is not empty


RemoteSettings.Dockerfile
```diff
@@ -27,9 +27,22 @@ COPY ./kinto-remote-settings ./kinto-remote-settings
...
+# We build the Kinto Admin assets at the specific
+# version specified in `kinto-admin/VERSION`.
+FROM node:20.11.0-slim as build-admin
+COPY /bin/update_and_install_system_packages.sh /opt
+RUN /opt/update_and_install_system_packages.sh \
+    ca-certificates\
+    curl
+COPY bin/build-kinto-admin.sh /opt/
+COPY kinto-admin/ /opt/kinto-admin/
+RUN cd /opt && ./build-kinto-admin.sh
```
grahamalama, line 39: - I think if we use the non-slim `node` base container, we can avoid downloading `curl`. Since we're copying the built assets from that container, it seems fine to use the non-slim version
- [Hadolint](https://github.com/hadolint/hadolint) (which I run locally) poked me about setting a `WORKDIR`. Doing so seemed like a nice change?

```suggestion
FROM node:20.11.0 as build-admin
WORKDIR /opt
COPY bin/build-kinto-admin.sh .
COPY kinto-admin/ kinto-admin/
RUN ./build-kinto-admin.sh
```

bin/build-kinto-admin.sh
```diff
@@ -0,0 +1,25 @@
...
+SRC_DIR="kinto-admin-${VERSION}"
+
+# download and unzip release
+curl -OL https://github.com/Kinto/kinto-admin/archive/refs/tags/${TAG}.tar.gz
+tar -xf $TARBALL_NAME && rm $TARBALL_NAME
+
+# build kinto-admin bundle
+pushd $SRC_DIR
+npm ci
+export SINGLE_SERVER=1
```
grahamalama, line 16: FYI: https://github.com/Kinto/kinto-admin/pull/3104#issuecomment-1893906526

grahamalama (approved):

PR #527, opened 2024-01-17
Upgrade kinto-admin to v2.1.2

leplatrem:


leplatrem:
See https://github.com/Kinto/kinto-admin/pull/3122

alexcottner:
As pointed out to me - that build parameter changing is a breaking change for us since we have downstream projects. Will be creating a v3.0.0 release after another PR is landed.

grahamalama:
@leplatrem going to close this PR and reopen for version 3.0.1

PR #549, opened 2024-02-06
Fixes #522 - Replacing selenium test with simple playwright test

### LINKED ISSUE #522: Investigate selenium test errors
We've had some recent inconsistent test runs with browser-tests where selenium will fail due to stale DOM elements. Initial investigations couldn't find a good solution so we have a `time.sleep(1)` line in the browser tests now. Locate and fix this the root cause.

alexcottner:
Fixes #522 by replacing selenium with playwright. These tests are simple and are meant as a general check that the kinto-admin UI exists, is accessible, and we can at least submit a review request. Detailed testing will be built out in the kinto-admin project.

Making these changes in `remote-settings` was a bit more complicated than `kinto`. Let me know if we'd like to see anything done differently.

### DIFF
--- a/IntegrationTests.Dockerfile
+++ b/IntegrationTests.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12.1-slim as build
+FROM python:3.12.1 as build
 
 ENV PIP_NO_CACHE_DIR=off \
     PIP_DISABLE_PIP_VERSION_CHECK=on \
@@ -17,7 +17,9 @@ WORKDIR /opt
 COPY pyproject.toml poetry.lock ./
 RUN $POETRY_HOME/bin/poetry install --only integration-tests --no-root
 
-FROM python:3.12.1-slim
+FROM python:3.12.1
+RUN apt update && apt upgrade -y
+
 ENV PATH="/opt/.venv/bin:$PATH" \
     PYTHONUNBUFFERED=1 \
     VIRTUAL_ENV=/opt/.venv \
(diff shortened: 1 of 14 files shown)


IntegrationTests.Dockerfile
```diff
@@ -17,7 +17,9 @@ WORKDIR /opt
 COPY pyproject.toml poetry.lock ./
 RUN $POETRY_HOME/bin/poetry install --only integration-tests --no-root
 
-FROM python:3.12.1-slim
+FROM python:3.12.1
+RUN apt update && apt upgrade -y
```
grahamalama, line 21: This should be handled in `update_and_install_system_packages.sh`, no?

IntegrationTests.Dockerfile
```diff
@@ -28,6 +30,9 @@ RUN /opt/update_and_install_system_packages.sh wget
 
 COPY --from=build $VIRTUAL_ENV $VIRTUAL_ENV
 
+RUN playwright install firefox
+RUN playwright install-deps
```
grahamalama, line 34: https://playwright.dev/docs/browsers#install-system-dependencies

Does `RUN playwright install --with-deps firefox` work?

tests/browser_test.py
```diff
@@ -1,168 +1,92 @@
...
+    page.get_by_text("Request review...").first.click()
+    page.get_by_placeholder("Comment...").last.fill("Review comment")
+    page.get_by_text("Request review").last.click()
+
+    # verify that we are in-progress for review
+    expect(page.locator(".bs-wizard-step.complete").first).to_contain_text("Work in progress")
+    expect(page.locator(".bs-wizard-step.active").first).to_contain_text("Waiting review")
+    expect(page.locator(".bs-wizard-step.disabled").first).to_contain_text("Approved")
+
+# TODO: reviewer test scenario - login to approve pending request
```
grahamalama, line 92: Is the plan to address this TODO in another PR? If so, let's turn it into an issue and remove this line

grahamalama (approved):
A few nits, but overall this is looking good. Goodbye Selenium 👋🏻

PR #557, opened 2024-02-08
Sign commit generated by github action for kinto-admin update

alexcottner:
Adjusting our github action workflow that updates kinto-admin. Making the request over the github API should result in a signed commit.

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -65,11 +65,26 @@ jobs:
             echo "pr_needed=false" >> $GITHUB_OUTPUT
           else
             git checkout -b "$BRANCH_NAME"
-            echo "$LATEST_RELEASE" > kinto-admin/VERSION        
-            git commit -am "Update Kinto Admin to $LATEST_RELEASE"
             git push origin $BRANCH_NAME
             echo "pr_needed=true" >> $GITHUB_OUTPUT
           fi
+      
+      - name: Commit change
+        if: ${{ steps.compare_versions.outputs.update_needed == 'true' && steps.create_branch.outputs.pr_needed == 'true'}}
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: updates/kinto-admin-${{ steps.latest_release_version.outputs.version }}
+          LATEST_RELEASE: ${{ steps.latest_release_version.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/scheduled.yaml
```diff
@@ -65,11 +65,26 @@ jobs:
...
+        if: ${{ steps.compare_versions.outputs.update_needed == 'true' && steps.create_branch.outputs.pr_needed == 'true'}}
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: updates/kinto-admin-${{ steps.latest_release_version.outputs.version }}
+          LATEST_RELEASE: ${{ steps.latest_release_version.outputs.version }}
+        run: |
+          echo "$LATEST_RELEASE" > kinto-admin/VERSION
+          SHA=$(git rev-parse $BRANCH_NAME:kinto-admin/VERSION)
+          CONTENT=$(base64 -i kinto-admin/VERSION)
+          gh api --method PUT /repos/:owner/:repo/contents/kinto-admin/VERSION \
```
grahamalama, line 82: Are the `:owner` and `:repo` placeholders magically interpolated somehow?

grahamalama (approved):

.github/workflows/scheduled.yaml
alexcottner: When running in an action, this context should be set for us automatically. You can set it locally by doing `gh repo set-default` inside a git repo and you will see something like:
```
~/Mozilla/remote-settings$ gh repo set-default 
Found only one known remote repo, mozilla/remote-settings on github.com.
✓ Set mozilla/remote-settings as the default repository for the current directory
```

If you're on a fork it'll ask you to pick which remote you want:
```
gh repo set-default 
This command sets the default remote repository to use when querying the
GitHub API for the locally cloned repository.

gh uses the default repository for things like:

 - viewing and creating pull requests
 - viewing and creating issues
 - viewing and creating releases
 - working with GitHub Actions
 - adding repository and environment secrets

? Which repository should be the default?  [Use arrows to move, type to filter]
  mozilla/remote-settings
> alexcottner/remote-settings
```

These merge fields fall into the *magic* category and I was going to hard-code them at first. But then this workflow isn't usable on forks, if the repo name ever changes, etc.

grahamalama:
For posterity: this is related to #529

PR #572, opened 2024-03-07
Standardizing browser tests between environments

alexcottner:
This PR standardizes the collection config across tested environments. This should allow us to have more confidence in automated releases, which will be the next step for us.

Also ran `make format` which touched a couple unexpected files.

I ran the test container produced by this against dev as an additional check.


leplatrem (approved):
Do we have assertion about `HTTP 400` responses on invalid posted data? 
If not, the collection JSON schema may not be necessary.

alexcottner:
It should only have an affect on environments where the collection doesn't exist (`if_not_exists` flag). Which could be in local testing or in CI pipelines.

PR #574, opened 2024-03-08
Switching admin-pull to use alpine:3 image

alexcottner:
Changing the docker image used to pull the admin files. Mainly doing this so dependabot stops [creating PR's](https://github.com/mozilla/remote-settings/pull/573) for it.

Also switched to wget while here. We will now error on 404's rather than following them to the pretty error page and failing when extracting the tar.

### DIFF
--- a/RemoteSettings.Dockerfile
+++ b/RemoteSettings.Dockerfile
@@ -29,7 +29,7 @@ RUN pip install ./kinto-remote-settings
 
 # We build the Kinto Admin assets at the specific
 # version specified in `kinto-admin/VERSION`.
-FROM node:21.6.2 as build-admin
+FROM debian:latest as get-admin
 WORKDIR /opt
 COPY bin/pull-kinto-admin.sh .
 COPY kinto-admin/ kinto-admin/
@@ -61,7 +61,7 @@ RUN chown 10001:10001 /app && \
 COPY --chown=app:app . .
 COPY --from=compile /opt/dogstatsd_plugin.so .
 
-COPY --from=build-admin /opt/kinto-admin/build $KINTO_ADMIN_ASSETS_PATH
+COPY --from=get-admin /opt/kinto-admin/build $KINTO_ADMIN_ASSETS_PATH
 
 # Generate local key pair to simplify running without Autograph out of the box (see `config/testing.ini`)
 RUN python -m kinto_remote_settings.signer.generate_keypair /app/ecdsa.private.pem /app/ecdsa.public.pem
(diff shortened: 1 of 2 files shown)


bin/pull-kinto-admin.sh
```diff
@@ -5,7 +5,7 @@ VERSION=$(cat `pwd`/kinto-admin/VERSION)
 TAG="v${VERSION}"
 
 # download and unzip release
-curl -OL https://github.com/Kinto/kinto-admin/releases/download/${TAG}/kinto-admin-release.tar
```
grahamalama, line 8: Can we avoid switching to `wget` by adding the `--fail` flag?

RemoteSettings.Dockerfile
```diff
@@ -29,7 +29,8 @@ RUN pip install ./kinto-remote-settings
 
 # We build the Kinto Admin assets at the specific
 # version specified in `kinto-admin/VERSION`.
-FROM node:21.6.2 as build-admin
+FROM debian:latest as get-admin
```
grahamalama, line 32: I think it's a bad idea to switch to a `latest` flag. Can we pin to `bookworm`, if not `bookworm-<date>`? It's always annoying when a CI pipeline magically starts failing and you're scrambling trying to figure out why. I'll take some Dependabot noise if it means avoiding silent failures.

I also wonder if we could use the `curl` image here, if that's all we need (well, plus `bash`). Don't know how that image compares with this one as far as how often new tags are published and size.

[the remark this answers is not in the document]
alexcottner: We could, I just went with debian as it's slow moving and figured it would be safe safe to point at `latest`

alexcottner: The curl images seem _very_ stripped down and won't have enough to work here. I've switched to `alpine:3` for now as that might be the easiest. But it doesn't include curl. Let me know if you'd rather still use curl as they don't include that by default.

I could also merge this into the `compile` container, but then we're re-pulling the UI every time something else changes.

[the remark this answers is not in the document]
alexcottner: Switched to alpine:3 for now as it appears to have enough standard packages to do everything.

bin/pull-kinto-admin.sh
alexcottner: Switched to alpine:3 but still using wget because it's included by default. Let me know if we're attached to curl though.

grahamalama: Not attached to curl as much as not having to install another dependency. If alpine comes with `wget`, that works for me 👍🏻

PR #600, opened 2024-04-18
Clarify usage of `_expected` querystring parameter

leplatrem:
@bendk do you find it clearer like this? Don't hesitate to ask more questions :) Thanks!


docs/client-specifications.rst
```diff
@@ -87,7 +87,7 @@ Returns the following response for the collection ``{cid}`` in the bucket ``{bid
 
 .. note::
 
-    The ``_expected={}`` querystring parameter is mandatory but can be set to ``0`` if unknown. See section below about cache busting.
+    The ``_expected={}`` querystring parameter is mandatory. Either you pass the current collection timestamp value obtained when polling for changes in order to bust the CDN cache, or you use a hard-coded value (eg. ``0``) and rely on the cache TTL. See section below about cache busting.
```
bendk, line 90:
```suggestion
    The ``_expected={}`` querystring parameter is mandatory. Either you pass the timestamp value obtained from the last fetch for the collection in order to bust the CDN cache, or you use a hard-coded value (eg. ``0``) and rely on the cache TTL. See section below about cache busting.
```

bendk (approved):
This is seems much clearer to me, but maybe it could be improved more by making it a bit more explicit (assuming that I'm understanding it correctly).  WDYT?

Approving in any case, because it's an improvement either way and I'm not sure that my suggestion is an improvement.

bendk:
One more question I had: how does `_expected` relate to `_since`?  Do I need to specify both at once?  If so, should they always be the same?

bendk:
I just read your comments in bugzilla and realize my understanding is totally wrong, since I wasn't really thinking about the push notification scenario.  Let me update my comments.

docs/client-specifications.rst
bendk, line 90: I'm not quite sure what "current collection timestamp" means.  Is it the timestamp of the last fetch?  The timestamp returned in the data for the last fetch?  The current time?  Whichever one it is, please spell out how it would work to bust the cache.

Can you lay out a scenario when you would want to bust the cache while polling?  In general, it seems like I wouldn't want to do that.

Also, how can we use the `_since` param in a cache-friendly manner?  If all clients did something naive like use the timestamp they last polled, then it seems like each request would be a cache-miss.

docs/client-specifications.rst
```diff
@@ -108,7 +108,7 @@ Returns the list of collections and their current timestamp.
 
 .. note::
 
-    The ``_expected={}`` querystring parameter is mandatory but can be set to ``0`` if unknown. See next section about cache busting.
+    The ``_expected={}`` querystring parameter is mandatory. Either you receive a Push notification from the server, and pass the timestamp value in order to bust the CDN cache, or you use a hard-coded value (eg. ``0``) and rely on the cache TTL. See section below about cache busting.
```
bendk, line 111: This one makes total sense to me.  Is this param is intended more for responding to push notifications than for polling?

alexcottner (approved):
Somehow lint is unhappy, but these changes seem to add good clarity

bendk:
That makes a lot of sense to me.  I'd love to see you list out some of those example use cases in the docs, but that doesn't need to happen in this PR.

PR #621, opened 2024-06-07
Fix details in ingestion job documentation

leplatrem:


alexcottner (approved):
Look like good clarifications

PR #631, opened 2024-06-27
ADR for client data bundling

alexcottner:
[Link to document for easy viewing](https://github.com/mozilla/remote-settings/blob/adr-003/docs/adr/adr_003_bundling_collection_data.md).

### DIFF
--- a/docs/adr/adr_003_bundling_collection_data.md
+++ b/docs/adr/adr_003_bundling_collection_data.md
@@ -1,0 +1,208 @@
+# Bundling Collection Data for Download
+
+
+* Status: proposed
+* Deciders: mleplatre, gbeckley, acottner
+* Date: June 27 2024
+
+Tracking issue(s):
+-  [The initial remote settings download is inefficient](https://bugzilla.mozilla.org/show_bug.cgi?id=1889617)
+
+
+## Context and Problem Statement
+When a new Firefox client comes online for the first time, it makes over 1,700 HTTP requests to download the initial remote settings data. These are mostly attachments from large collections (ex: `security-state/intermediate, blocklists/certificates`). Opening this many HTTP connections to download small files is not as efficient compared to downloading a few larger files.
+1. It’s a poor user experience if the UI locks up while trying to maintain this many connections
+2. It causes excess power usage and [battery drain](https://bugzilla.mozilla.org/show_bug.cgi?id=1889617) on laptop and mobile devices by keeping modems/wifi devices running for longer than they need to 
+
+If we can bundle these downloads together by collection, we could reduce the number of requests for certain collections from almost thousands to a few. This should reduce the overall time required to download the data (less time establishing connections), which will reduce power usage and improve the user experience.
+
+
+### Data Bundling
+We distinguish two parts for this data bundling:
+
+ - A bundle for all records data for new profiles to fetch all collections data in one request instead of 50 (~6MB uncompressed); 
+ - A bundle of all attachments data per collection (opt-in)
+
+Looking at existing use-cases, we can observe that bundling attachments may not be equally relevant depending on the number of records and their attachment size:
+ - `security-state/intermediates` collection: 1720 records for 3MB in total. Lots of tiny .pem files required as soon as possible after first startup for security features: __bundling very beneficial__
+ - `fakespot-suggest-products` collection: 300 records with a few MB per attachment (eg. 5MB). Since the total size is huge: bundling brings __little advantage__
+ - Machine learning collections: few records with big attachments (eg. 20MB): almost no advantage. 
+ - `Translations-models` collection: 150 records for 1.2GB in total. Clients only download a few pairs when the user enables the feature for certain languages: __bundling is not feasible__
+
+That’s why it would make sense for attachments bundling to be opt-in for stakeholders.
+
+There is already a collection metadata attribute to mark attachments as enabled and/or required. The flag would just consist in adding a new field attachment.bundle: true|false.
+
+Also, since [usage quotas](https://docs.kinto-storage.org/en/stable/api/1.x/quotas.html) are not enabled on the Remote Settings server, consumers can end up publishing huge files and bloat attachments bundles, we may want to have a server side configuration setting to fix a size limit for which the bundles won’t be built. For example, even when opted-in, if the target bundle exceeds 20MB, it won’t be built and published. Clients should thus be smart and fail soft if a bundle is not available.
+
+
+### Bundle Formats
+
+- Tar - When gzipped (automatically by cdn dynamic compression), we can expect to save around 43% size for text content. Probably not much for images.
+    - Pros: very easy to work with server side and client side. Compression and decompression are essentially free at the transport layer.
+    - Cons: doesn’t save as much as lzma
+- LZMA (7z) - Reduces text content up to 70%.
+    - Pros: Very compressed
+    - Cons: More compute intense to compress and decompress. Likely need to ship another library to decompress.
+- Zip - Reduces text content up to 17%.
+    - Pros: Easy to work with and probably decompress client side.
+    - Cons: Not much compression. Probably doesn’t make sense over tar. Need to decompress after receiving.
+
+### Client Implementation
+This ADR mostly covers the server parts, since the client implementation to fetch and leverage bundling does not have many different possible approaches.
+
+For the collections records bundle, clients with empty profiles will simply pull it on first synchronization:
+- Pull the latest records bundle
+- Extract it and store data into local IndexedDB
+- Fetch latest timestamps from the server (like currently using `Utils.fetchLatestChanges()`)
+- Synchronize potential outdated collections, and verify signatures of the ones that are up-to-date, to guarantee integrity/authenticity of the downloaded bundle
+
+For the attachments, bundling would be opt-in, and leveraged explicitly by consumers in their code. Clients would have a new method `client.attachments.cacheAll()` that would:
+- Pull the attachment bundle for the current collection timestamp
+- Resume a previously failed download
+- Be resilient to missing bundles (eg. not yet available or exceeding configured server size limits)
+- Extract it on disk (TODO: find example of zip extraction in Gecko)
+- Put files contents in attachments cache ([source](https://searchfox.org/mozilla-central/rev/8ec3cc0472ad4f51b254728d024b696eaba82ba0/services/settings/Attachments.sys.mjs#259-261))
+- Delete temporary content
+
+After that, any call to `client.attachments.download()` would return the locally cached content as it does currently if the attachment has been downloaded previously. It also checks its hash using sha256 ([source](https://searchfox.org/mozilla-central/rev/b11735b86bb4d416c918e2b2413456561beff50c/services/settings/Attachments.sys.mjs#449-470)), which saves us from having to add content signature on the attachment bundle.
+
+In future calls of `download()` on records whose attachment has been modified on the server, the client would fetch them individually as it does with the current code. In other words, in this ADR, we don’t address building “delta” or “diff” bundles for clients that already have local data.
+
+Consumers would choose to explicitly use this new method in their code, based on their use-case. And we (SysEng) could easily modify existing code to rewrite the code of some use-cases that rely on massive parallel fetch (eg. replace with a call to `cacheAll()`).
+
+Requirements
+- Total amount of time to download, extract, and process, must not take longer than fetching attachments individually
+- Handling the bundle in the client should not bloat the main thread or impact user experience in general
+- Feature must not get in the way of legacy clients
+- Implementation should be resilient to race conditions on attachment downloads. Unpacking the bundle can overwrite attachments that had been previously downloaded individually, and individual downloads can overwrite unbundled ones
+- Bundle format should be easily readable by client (third-party libraries are hard to ship in Gecko)
+
+References
+- [The intermediates download code](https://searchfox.org/mozilla-central/rev/8ec3cc0472ad4f51b254728d024b696eaba82ba0/security/manager/ssl/RemoteSecuritySettings.sys.mjs#335-365) that fetches attachments in bulk, and stores state in records
+- [Download of crlite files](https://searchfox.org/mozilla-central/rev/8ec3cc0472ad4f51b254728d024b696eaba82ba0/security/manager/ssl/RemoteSecuritySettings.sys.mjs#620-637)
+- [Download of all regions files](https://searchfox.org/mozilla-central/rev/8ec3cc0472ad4f51b254728d024b696eaba82ba0/security/manager/ssl/RemoteSecuritySettings.sys.mjs#335-365)
+- [Download of all addons suggestions files](https://searchfox.org/mozilla-central/rev/8ec3cc0472ad4f51b254728d024b696eaba82ba0/browser/components/urlbar/private/AddonSuggestions.sys.mjs#87-101)
+
+### Metrics To Measure Before/After
+CPU usage, bandwidth, usage for full synchronization. Power and performance profiles on different clients.
+    - Mac Desktop
+    - Linux Desktop
+    - Windows Desktop
+    - Android Mobile
+- iOS Mobile (does it use remote-settings?)
+- How much collections are compressed
+- Total duration of synchronization on fresh profiles
+
+
+## Non Goals
+- Support diffs or delta bundles for clients that already have local data
+- Reduce client performance (ex: we shouldn't increase power usage for clients)
+
+## Assumptions
+- We will find a way for Gecko clients to extract at least one archive format (very likely using third-party library)
+
+## Decision Drivers
+- Keep costs low
+    - With millions of clients, the cost of shipping data is always a concern
+- Server complexity & performance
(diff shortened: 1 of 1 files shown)


docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,209 @@
...
+    -Monitoring and operating should remain straightforward (ex: limit tech and deployment complexity if possible)
+    -Performance of the service should not be affected
+    - The code in place to support this feature should not involve complex components, libraries, or architecture (eg. to support resumable downloads and range requests)
+- Long-term Maintenance
+- Clients complexity
+    - How complicated it is to know which bundle URL to pull
+
+## Considered Options
+1. Build bundles on attachment upload/removal
+2. Build bundles on Cloud storage event
```
grahamalama, line 119: If I understand correctly, both options 1 and 2 (re)build bundles on attachment upload/removal. Option 1 handles this in the request / response cycle, and option 2 offloads the work to a cloud function. 

If that's the case, maybe we can include `Build bundles on attachment upload/removal` in both option names. For example:

```suggestion
1. Synchronously build bundles on attachment upload/removal (in request / response cycle)
2. Aynchronously build bundles on attachment upload/removal (via Cloud Storage event)
```

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+## Non Goals
+- Support diffs or delta bundles for clients that already have local data
+- Reduce client performance (ex: we shouldn't increase power usage for clients)
+
+## Assumptions
+- We will find a way for Gecko clients to extract at least one archive format (very likely using third-party library)
+
+## Decision Drivers
```
grahamalama, line 106: We mention whether the solution is generic (i.e. could be included in `kinto-attachment`) or Remote Settings specific a few times in the options below. Should we mention that as a decision driver here?

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Considered Options
+1. Build bundles on attachment upload/removal
+2. Build bundles on Cloud storage event
+3. Build bundles on RS changes approval
+4. Add new API endpoint to build or serve last bundle
+5. Build bundles on schedule
+
+## Decision Outcome
+Chosen option: [Option 5](#Option-5) because it is easy to reason about, trivial to implement, does not affect user experience, and would not increase the overall service complexity.
```
grahamalama, line 125: One thing that isn't immediately clear (to me) about some of the options and the resulting decision is how we weigh some pros and cons against each other.

For example, we call option 5 "trivial" to implement, but there are some aspects of it that make it seem less-trivial than Option 1. But we're willing to take on some non-trivialness for the performance / UX improvements we'd gain, which makes sense. Maybe we can expand this decision outcome to mention that.

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,209 @@
...
+
+We would have to duplicate some of the `kinto-attachment` configuration values (eg. `kinto.attachment.gcloud.bucket_name` and `kinto.attachment.folder`) in order to store the bundles along attachments (ie. in the right folder `{bucket id}/{collections id}`), so that the final bundle URL remains straightforward for clients (eg. `capabilities.attachments.base_url` + `bucket_id/collection_id /bundle.zip`)
+
+The clients would pull the records bundle on first sync, and the attachments bundle URL when `attachments.cacheAll()` is called. 
+- Good, because ultra simple
+- Good, because low cost
+- Good, because monitored as other remote settings cronjobs
+- Good, because it works easily for the preview bucket
+- Not so good, because we would introduce some coupling with kinto-attachment config (unless we mount the same RS server .ini config in the container and read conf from it) 
+- Not so good, because if we don’t couple it with the push timestamp cronjob, there could be a window of time where the bundle for a particular timestamp is not yet available
```
grahamalama, line 205: We call this option "trivial" to implement, but these concerns don't exactly seem trivial.

grahamalama (commented):
Looks sound overall. A few nit comments / suggestions inline.

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,209 @@
...
+* Status: proposed
+* Deciders: mleplatre, gbeckley, acottner
+* Date: June 27 2024
+
+Tracking issue(s):
+-  [The initial remote settings download is inefficient](https://bugzilla.mozilla.org/show_bug.cgi?id=1889617)
+
+
+## Context and Problem Statement
+When a new Firefox client comes online for the first time, it makes over 1,700 HTTP requests to download the initial remote settings data. These are mostly attachments from large collections (ex: `security-state/intermediate, blocklists/certificates`). Opening this many HTTP connections to download small files is not as efficient compared to downloading a few larger files.
```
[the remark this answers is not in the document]
alexcottner: Only if they get too far out of sync. We will have a follow up down the road to implement minimal binary replication that would help established clients (potentially by a lot).

docs/adr/adr_003_bundling_collection_data.md
alexcottner: Adjusting these to be a list and mentioned that

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Bundle Formats
+After some research and testing, we've decided to go with Zip compression. Zip is very easy to implement and allows for good flexibility moving forward. We could change our minds or add other bundle types in the future with very little effort.
+
+- Zip - Reduces text content up to 17%.
+    - Pros: Easy to work with. Easy to decompress client side.
+    - Cons: Not as much compression as lzma. Need to decompress after receiving.
+- Tar - When gzipped (automatically by cdn dynamic compression), we can expect to save around 43% size for text content. Probably not much for images.
+    - Pros: very easy to work with server side and client side. Compression and decompression should be free at the transport layer.
+    - Cons: doesn’t save as much as lzma.
```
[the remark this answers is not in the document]
alexcottner: Added more notes. I think the main things are:
1. Dynamic compression requires the tar file to be < 10MB
2. We have built-in zip extraction thanks to `ZipReader` in firefox
3. It's very easy for us to change this later

docs/adr/adr_003_bundling_collection_data.md
alexcottner: > because we would introduce some coupling with kinto-attachment config

I think this is mainly us pulling the same environment values, which is hopefully not too big of an issue.

> if we don’t couple it with the push timestamp cronjob, there could be a window of time where the bundle for a particular timestamp is not yet available

This is the normal CDN expiration concern, right? Maybe I'm minimizing this concern.  I think the negative scenario here is a client pulls yesterday's bundle and then has to pull some deltas for a few minutes. But if they had pulled the bundle just before publishing it's the same scenario. That doesn't seem too bad to me.

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+1. It’s a poor user experience if the UI locks up while trying to maintain this many connections
+2. It causes excess power usage and [battery drain](https://bugzilla.mozilla.org/show_bug.cgi?id=1889617) on laptop and mobile devices by keeping modems/wifi devices running for longer than they need to 
+
+If we can bundle these downloads together by collection, we could reduce the number of requests for certain collections from almost thousands to a few. This should reduce the overall time required to download the data (less time establishing connections), which will reduce power usage and improve the user experience.
+
+Note: This improvement is primarily targeting new clients, or clients that are very out of date. Clients that are established and updating daily will not see improvement from this. However, we will be working on new binary replication logic that will help established clients in the future.
```
leplatrem, line 19: I don't think we should put this here: `However, we will be working on new binary replication logic that will help established clients in the future.`
Because it may never be true

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+
+Looking at existing use-cases, we can observe that bundling attachments may not be equally relevant depending on the number of records and their attachment size:
+ - `security-state/intermediates` collection: 1720 records for 3MB in total. Lots of tiny .pem files required as soon as possible after first startup for security features: __bundling very beneficial__
+ - `fakespot-suggest-products` collection: 300 records with a few MB per attachment (eg. 5MB). Since the total size is huge: bundling brings __little advantage__
+ - Machine learning collections: few records with big attachments (eg. 20MB): almost no advantage. 
+ - `Translations-models` collection: 150 records for 1.2GB in total. Clients only download a few pairs when the user enables the feature for certain languages: __bundling is not feasible__
+
+That’s why it would make sense for attachments bundling to be opt-in for stakeholders.
+
+There is already a collection metadata attribute to mark attachments as enabled and/or required. The flag would just consist in adding a new field attachment.bundle: true|false.
```
leplatrem, line 35:
```suggestion
There is already a collection metadata attribute to mark attachments as enabled and/or required. The flag would just consist in adding a new field `attachment.bundle: true|false`.
```

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+    - Cons: 
+        - No existing interface in gecko.
+        - Requires dynamic compression.
+- LZMA (7z) - Reduces text content up to 70%.
+    - Pros: 
+        - Very compressed
+    - Cons: 
+        - More compute intense to compress and decompress
+        - Likely need to ship another library to decompress
+        - Need to decompress after receiving
```
leplatrem, line 66: We said `Double compression if dynamic compression is used` for Zip

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,209 @@
...
+
+
+### Client Implementation
+This ADR mostly covers the server parts, since the client implementation to fetch and leverage bundling does not have many different possible approaches.
+
+For the collections records bundle, clients with empty profiles will simply pull it on first synchronization:
+- Pull the latest records bundle
+- Extract it and store data into local IndexedDB
+- Fetch latest timestamps from the server (like currently using `Utils.fetchLatestChanges()`)
+- Synchronize potential outdated collections, and verify signatures of the ones that are up-to-date, to guarantee integrity/authenticity of the downloaded bundle
```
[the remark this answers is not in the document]
leplatrem: Having content signatures on the bundles could have some benefits indeed. 
However, it would require the cronjob to have access to Autograph.

With the proposed approach, the bundle itself is not verified but:
- we verify the signature of the local data for each collection after its extraction
- the signature of the cached attachment is verified when `client.attachments.download()` is called

So, if the bundle is corrupted, the local extracted data and attachments would be erased and refetched from server if signature verification fails.

We could have a `.md5sum` file though, just to verify the basic integrity of the downloaded file before trying to extract it.

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+
+For the attachments, bundling would be opt-in, and leveraged explicitly by consumers in their code. Clients would have a new method `client.attachments.cacheAll()` that would:
+- Pull the attachment bundle for the current collection timestamp
+- Be resilient to missing bundles (eg. not yet available or exceeding configured server size limits)
+- Extract it on disk (TODO: find example of zip extraction in Gecko)
```
leplatrem, line 81:
```suggestion
- Extract it on disk
```

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+- Android Mobile
+- iOS Mobile (does it use remote-settings?)
+
+
+## Non Goals
+- Support diffs or delta bundles for clients that already have local data
+- Reduce client performance (ex: we shouldn't increase power usage for clients)
+
+## Assumptions
+- We will find a way for Gecko clients to extract at least one archive format (very likely using third-party library)
```
leplatrem, line 120: This is obsolete now that we can extract Zip content

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+More alternatively, we could use event-driven architecture. The API would submit a request for the bundle to be built to a queue. A serverless function would process the request and return the response. In order to prevent multiple to build the same bundle multiple times, parallelism would have to be restricted, or we’d need to use a queue that supports partitioning based on collection name (or similar).
+- Good, because it all happens in one endpoint and easy to reason about
+- Good, because uploading/removing attachments or approving changes would not take longer
+- Bad, because a new specific container that has to be monitored, troubleshot, and deployed
+- Bad, because we have to introduce some complexity in order to prevent building several bundles in parallel
+
+### Option 5 - Build bundles on schedule
+A scheduled job iterates through the collections with the `attachment.bundle == true` flag and operates with the following logic:
+- If there is a bundle for the current collection timestamp, nothing to do.
+- If there isn’t, pull all attachments from storage, build a bundle with all current attachments, store it at `{folder}/attachments.zip` in cloud storage
```
leplatrem, line 209:
```suggestion
- If there isn’t, pull all attachments from storage, build a bundle with all current attachments, store it at `bundles/{bucket-id}--{collection-id}.zip` in cloud storage
```

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+- (repeat for each bucket, including preview buckets)
+
+And creates a records bundle with all collections data (~6MB) for each bucket (`{folder}/changesets.zip`).
```
leplatrem, line 212:
```suggestion
And creates a records bundle with all collections data (~6MB) for each bucket (`bundles/changesets.zip`).
```

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -0,0 +1,230 @@
...
+
+*Note: the job that is currently in charge of publishing the highest timestamp to the Push notification server, could first execute this attachment bundling job, in order to make sure bundles will be available when clients will pull changesets.*
+
+This job would require GCloud credentials (Service Accounts credentials JSON file) and unlike `kinto-attachment` would be vendor-specific (ie. GCP Cloud storage using the `google-cloud-storage` Python library).
+
+We would have to duplicate some of the `kinto-attachment` configuration values (eg. `kinto.attachment.gcloud.bucket_name` and `kinto.attachment.folder`) in order to store the bundles along attachments (ie. in the right folder `{bucket id}/{collections id}`), so that the final bundle URL remains straightforward for clients (eg. `capabilities.attachments.base_url` + `bucket_id/collection_id /bundle.zip`)
```
leplatrem, line 218:
```suggestion
We would have to duplicate some of the `kinto-attachment` configuration values (eg. `kinto.attachment.gcloud.bucket_name` and `kinto.attachment.folder`) so that the final bundle URL remains straightforward for clients (eg. `capabilities.attachments.base_url` + `bundles/{bucket-id}--{collection-id}.zip`)
```

docs/adr/adr_003_bundling_collection_data.md
leplatrem: > I think this is mainly us pulling the same environment values, which is hopefully not too big of an issue.

Yes, there is a little bit of duplication, but this remains trivial.

> if we don’t couple it with the push timestamp cronjob, there could be a window of time where the bundle for a particular timestamp is not yet available


Either we run the cronjob jobs together: ` python build_bundles.py && python sync_megaphone.py` (trivial)

Orwe accept that the few clients that will start a new profile right before the bundles are built, will pull an outdated bundle, and will have a few collections to synchronize after the extraction. (minor consequences)

We could add some details here if necessary

leplatrem (approved):

docs/adr/adr_003_bundling_collection_data.md
alexcottner: That's fair, priorities could shift.

docs/adr/adr_003_bundling_collection_data.md
alexcottner: Added to this section as well.

docs/adr/adr_003_bundling_collection_data.md
alexcottner: Yup, updating to none

docs/adr/adr_003_bundling_collection_data.md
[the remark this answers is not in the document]
alexcottner: I suggest that we put md5 sums as a possible future item. They could be helpful for resumable downloads to know if the bundle changed.

grahamalama (approved):

docs/adr/adr_003_bundling_collection_data.md
```diff
@@ -1,7 +1,7 @@
 # Bundling Collection Data for Download
 
 
-* Status: building
+* Status: landed, target version 130
```
leplatrem, line 4: You called the curse 😜 

I think having it as `accepted` or `planned` is enough

alexcottner: 10-4.

PR #633, opened 2024-07-03
Always copy 'schema' and 'attachment' attributes to preview and final collections

leplatrem:
Change the original code contributed here: https://github.com/Kinto/kinto-signer/pull/163

I don't remember why (8 years ago) we had this condition for fields: `(if not set)`

Without the change of this PR, we can observe that the destination attributes are never updated.
Example:
https://firefox.settings.services.mozilla.com/v1/buckets/main-preview/collections/search-telemetry-v2
Does not contain https://github.com/mozilla-services/remote-settings-permissions/commit/a22956c3931cec03b130770708288e340128d9a3

This is required for our work on bundles https://github.com/mozilla-services/remote-settings-lambdas/pull/1474 so that the `attachment.bundle: true` flag is copied to preview and main collections.


alexcottner (approved):

alexcottner:
simple, elegant

PR #695, opened 2024-11-06
Allow any file extension for attachments

willdurand:
This is needed to be able to submit attachments for AMO (addons-bloomfilters). Apparently, this is the config on prod too.


leplatrem (approved):
👌 

Merci!

PR #725, opened 2024-12-20
Sort entries by `-last_modified` in monitor/changes changeset endpoint

leplatrem:
Mimic behaviour of `monitor/changes` records endpoint

The list of collections in the DevTools will now be sorted by `-last_modified` since the order of the list comes from the `RemoteSettings.inspect()` function ([source](https://searchfox.org/mozilla-central/rev/f53c09a22edc700ab1a9eaaf4da0f0dd9f11bff3/services/settings/remote-settings.sys.mjs#604-638))

### DIFF
--- a/kinto-remote-settings/src/kinto_remote_settings/changes/views.py
+++ b/kinto-remote-settings/src/kinto_remote_settings/changes/views.py
@@ -311,8 +311,13 @@ def get_changeset(request):
         last_modified = (
             records_timestamp  # The collection 'monitor/changes' is virtual.
         )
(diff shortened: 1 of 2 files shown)


leplatrem:
> If we modify how changeset sorts things, there might be a few other impacts to think about. Like the remote-settings-devtools users. But that would be a clean solution if we make everything use changeset and it sorts things in an expected way.

You're right, clients would be instantiated by order of modified (collections with oldest last change would end up at the bottom). I think it's fine :) but let's look at this next year :)

leplatrem:
See https://github.com/mozilla-services/telescope/pull/1523

PR #729, opened 2025-01-03
Add ADR about server side filtering

leplatrem:


docs/adr/adr_004_server_side_filtering.md
```diff
@@ -0,0 +1,142 @@
...
+    "id": "shop-A",
+    "regions": ["europe", "asia", "north-america"],
+},
+{
+    "id": "shop-B",
+    "regions": ["europe", "africa"],
+}
+```
+
+Then different collections can be populated using this field in [a querystring filter](https://docs.kinto-storage.org/en/latest/api/1.x/filtering.html#comparison):
```
ncloudioj, line 81: Question: is it possible to support multiple dimensions for side collections? For example:
```
{
    "id": "ship-A",
    "regions": ["europe", "asia", "north-america"],
    "locales": ["en", "fr", "de", "es"],
}

* `shops-europe-en`: `?contains_regions=["europe"]&contains_locales=["en"]`
* `shops-enrope-fr`: `?contains_regions=["europe"]&contains_locales=["fr"]`
....
```

ncloudioj:
@leplatrem Thanks for putting this up together!

I agree with you that "option 2" appears to be the most reasonable choice at the moment for its low complexity & cost and its ability to meet our present business needs.

Longer term, from a consumer's perspective, I'd be curious about the viability of "option 4" specifically regarding how to handle the increased signing workload for server-side filtering. If the scalability problem is intrinsically hard to solve, would it be reasonable for us to opt out content signing for certain use cases to work around that challenging? An interesting observation we have made from building user-facing content serving features in Firefox over the past years, we've been gradually shifting our mental model from error prevention to error tolerance & rapid responding. Granted it will lose certain security protections w/o content signing, though that lose could be totally acceptable if we have other mechanisms in place for emergency/incident handling. As an RS consumer, having such flexibility seems quite useful and compelling in certain cases especially if it'd unlock other capabilities.

docs/adr/adr_004_server_side_filtering.md
leplatrem: Yes it is possible, will update the doc :) 

All conditions are a conjunct ("AND"), and we don't have a way to do disjunctions though ("OR") ([issue opened since 2016](https://github.com/Kinto/kinto/issues/639) 🤣)

ncloudioj: Very nice! Conjunctions are all we need :)

docs/adr/adr_004_server_side_filtering.md
```diff
@@ -0,0 +1,146 @@
...
+With this solution, the main collection is maintained, and several "*side collections*" are created. As with *Option 1*, the clients are able to pick which collection to pull from.
+
+Unlike *Option 1*, the additional collections are read-only. Editors publish data in the main collection, and a scheduled job will copy the records at regular intervals to the side collections using filters.
+
+The [`backport_records` cronjob](https://github.com/mozilla-services/remote-settings-lambdas?tab=readme-ov-file#backport_records) copies records from one collection to another and signs it. It can take querystring filters as parameters in order to only copy of subset of the source.
+
+For example, if records have `regions` and `locales` fields:
+
+```
+{
```
alexcottner, line 71: What could collection config look like in this example to know how the records should be pivoted?

docs/adr/adr_004_server_side_filtering.md
```diff
@@ -0,0 +1,146 @@
...
+
+The clients would pull the subset of data, and use the specific signature to verify the integrity.
+
+- **Complexity**: Medium-High. We introduce some coupling between server configuration and client behaviour.
+- **User experience**: Good, users edit a single collection.
+- **Cost of implementation**: Medium-High. Server side code is relatively straightforward, but the different clients implementations have to be modified to support this new type of metadata.
+- **Cost of operation**:  Low. Clients download only the subset of data, saving bandwidth, and signing datasets on each publication is relatively cheap.
+
+
+### Option 4 - Implement dynamic signatures
```
alexcottner, line 139: We could have a hybrid approach here. Where we create the result on the first request and cache it. However if many clients make the request at the same time, we would need to implement a queue so we don't do the same work for each of them.

alexcottner:
I think option 2 is our best value without major changes.

Option 4 feels like it's not quite complete but could be great. We could setup a custom sync endpoint where the client provides info on it's current state, and a signed bundle is returned with exactly what it needs to be brought up to date. Something similar to couchDB's _changes endpoint, but with a signature attached. But we'd need some additional infrastructure to prevent competing requests from being processed at the same time.

docs/adr/adr_004_server_side_filtering.md
```diff
@@ -0,0 +1,146 @@
...
+And since the signature is computed on the server side for the whole dataset, clients are obliged to also download the same whole dataset locally in order to verify the data integrity.
+
+In this document, we explore possible solutions to overcome this limitation.
+
+## Decision Drivers
+
+In order to choose our solution we considered the following criteria:
+
+- **Complexity**: Low → High: how complex is the solution
+- **User experience**: Bad → Good: how adapted to users is the solution
```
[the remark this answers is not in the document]
leplatrem: Do you mean *developer experience*? Or *client API complexity*?

I could be wrong but I think Firefox users should not be impacted by this internal implementation details. Do you have something concrete in mind?

docs/adr/adr_004_server_side_filtering.md
leplatrem: What do you mean with *records should be pivoted*? 
How would this `pivotBy` field be used?


We could add a few lines below to show the client code required to leverage this option.
It would look like this:

```js
ChromeUtils.defineESModuleGetters(lazy, {
  Region: "resource://gre/modules/Region.sys.mjs",
});

const region = lazy.Region.home || "default";
const locale = Services.locale.appLocaleAsBCP47.substr(0, 2);

client = RemoteSettings(`shops-${region}-${locale}`);
```

leplatrem:
> Option 4 feels like it's not quite complete but could be great. 

Should I add more details?  

> We could setup a custom sync endpoint where the client provides info on it's current state, and a signed bundle is returned with exactly what it needs to be brought up to date. 

This could be one idea of improvement, but I feel that I does not relate directly to server side filtering.

docs/adr/adr_004_server_side_filtering.md
alexcottner: Sorry, maybe this was a bad assumption on my part. I assumed there would be a generic automated process that looks for these collections that need to create/update the read-only virtual collections. And that process would need to know how to disperse those records to the virtual collections when it runs.

In my head this is basically a generic ETL process that takes the records from the collection, and applies some "pivot table" like logic to disperse these records out.

I realize now there could be multiple processes that have their own custom logic in them, one for each client or collection. But I still like the idea of having a paved road for a generic process where we could list which properties we want to create collections for and it would just run with it.

leplatrem: If I understand correctly, you were looking for a generic way to create and populate all the sub-collections, and on the client side "route" them to the right sub-collection.

With the proposed solution here, the sub-collections are hardcoded in the cronjobs manifests, and the clients have hardcoded logics to pick the proper sub-collection.

alexcottner: Yes, exactly that. But you were thinking about storing that config in the cronjobs. I guess the end result is the same and this is down to small implementation details. Although if we put the config in the collection properties, admins could see it (and even play with it in dev) :thinking: . That could allow for some self-service.

Either way, we don't need to decide it here.

alexcottner:
> This could be one idea of improvement, but I feel that I does not relate directly to server side filtering.

You're right, that is a more generic solution and doesn't apply directly to server side filtering. We can leave that out for now.

docs/adr/adr_004_server_side_filtering.md
leplatrem: The generic/automatic/self-serve solution is neat, but unlike the proposed option, it requires some (minor) development. 
Shall we add another entry in this doc and evaluate?
The configuration of this "pivot" would require 2 aspects to be configured: the list of fields, and the pattern of the side collections.
For example `pivotBy: ["region", "locale"]`, and `targetPattern: "shop-{region}-{locale}"`.
I think it would still require us to create the side collections manually, and the cronjob would just "ignore" the records if they no existing collection matches (eg. if there's no `shop-oceania-de` collection, do not put this records anywhere)

alexcottner:
Because I  like diagrams, does this look right?
jpg_diagram

alexcottner (approved):
Loving it

alexcottner:
> > Because I like diagrams, does this look right?
> 
> Yes! I would define _virtual_ here, because with Option2 they are real/concrete collections

Yeah virtual is the wrong word here. Maybe "staging" or "calculated" or "pre-filtered" or something?

leplatrem:
> > > Because I like diagrams, does this look right?
> > 
> > 
> > Yes! I would define _virtual_ here, because with Option2 they are real/concrete collections
> 
> Yeah virtual is the wrong word here. Maybe "staging" or "calculated" or "pre-filtered" or something?

Calculated or computed seems better 👌

docs/adr/adr_004_server_side_filtering.md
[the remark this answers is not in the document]
leplatrem: All of this makes sense.
I rephrased some parts but didn't find justified to create a criteria for each of these. Let me know if you don't agree, I'd be happy to adjust more until it's good enough

leplatrem:
> Loving it

Shall I merge? Did you want to add a diagram or more options? We can also always iterate...

alexcottner:
Added the diagram to the decision outcome section. If this looks good, let's merge it and start getting this built out.

PR #734, opened 2025-01-14
Mention server error responses in clients specifications

leplatrem:


docs/client-specifications.rst
```diff
@@ -153,6 +153,39 @@ Examples:
 * `fetchLatestChanges() in Gecko <https://searchfox.org/mozilla-central/rev/1f27a4022f9f1269d897526c1c892a57743e650c/services/settings/Utils.sys.mjs#376-457>`_
 
 
+Server Error Responses
+''''''''''''''''''''''
+
+Every response is JSON.
+
+If the HTTP status is not OK (<200 or >=400), the response contains a JSON mapping, with the following attributes:
```
alexcottner, line 161: I don't think we return any response codes < 200? I don't think pyramid supports them.

alexcottner (approved):
Approved with a nit comment.

PR #753, opened 2025-02-07
Enable Github Actions updates via Dependabot

leplatrem:


alexcottner (approved):
Didn't realize we could do this

PR #758, opened 2025-02-10
Fix #755: Import lambdas into this repo

### LINKED ISSUE #755: Merge `mozilla-services/remote-settings-lambdas` into this repo
Having the Remote Settings cronjobs code in the `mozilla/remote-settings` repo would make sense.

We would have one less repo to maintain :) 

As far as I understand, the only consequence would be that they would be redeployed each time we tag a new RS version, and their version aligned with the RS version.

leplatrem:
Fix #755: Import lambdas into this repo

This will require some changes in the deploy repo. Instead of assuming that the lambdas container is built from the `Dockerfile` at the root of the `mozilla-services/remote-settings-lambdas` repo, it will now have to be built from the `Cronjobs.Dockerfile` file of `mozilla/remote-settings repo.

See also https://github.com/mozilla-sre-deploy/deploy-remote-settings-cronjobs/pull/17


cronjobs/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[tool.poetry]
+name = "cronjobs"
+version = "0"
+description = "Remote Settings core cronjobs"
+authors = ["Contactless <postmaster@localhost>"]
+
+[tool.poetry.dependencies]
+python = ">=3.11, <3.14"
```
alexcottner, line 8: Nice, upgraded from the requirements files

alexcottner (approved):

PR #811, opened 2025-03-21
Fix #808: reflect divergence of clients support in JEXL docs

### LINKED ISSUE #808: JEXL filtering documentation does not reflect multiplatform clients
Most fields on Desktop won't be available on mobile (application-services).

See application context fields on a-s on [Android](https://phabricator.services.mozilla.com/D241066) and [iOS](https://github.com/mozilla-mobile/firefox-ios/pull/25287)

leplatrem:
Fix #808
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1944609

In this PR, I split the whole list of JEXL filters into 3 sections:
- available in both desktop and application-services
- only application-services (stuff apparently copied from Nimbus)
- only desktop (most of the previous docs)

This way is much clearer for authors that certain JEXL filters won't work everywhere

### DIFF
--- a/docs/target-filters.rst
+++ b/docs/target-filters.rst
@@ -26,8 +26,8 @@ In order to restrict a setting to a particular audience, just write the proper f
     {
         id: "68b19efa-1067-401b-b1c1-8d7b4263bb86",
         last_modified: 1531762863373,
-        title: "Only for users of 57",
-        filter_expression: "env.version == 57"
+        title: "Not for Android users",
+        filter_expression: "env.appinfo.OS != 'Android'"
     }
 
 When calling ``RemoteSettings("key").get()`` or listening to ``sync`` events, you will only see the settings entries whose ``filter_expression`` resolved to a truthy value (and those who don't have any as by default).
@@ -111,6 +111,7 @@ additional arguments passed in the expression:
 .. code-block:: javascript
 
    '1980-01-07'|date // == a date object
+   env.version|versionCompare("110.0a1") >= 0
 
 .. _JEXL Readme: https://github.com/TechnologyAdvice/Jexl#jexl---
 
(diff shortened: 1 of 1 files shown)


alexcottner (approved):

docs/target-filters.rst
```diff
@@ -143,19 +144,222 @@ filter expressions.
...
+
+.. important::
+
+   **How to add a new field?**
+
+   1. Put Remote Settings stakeholders in the loop (would allow to avoid disparities like casing, etc.)
+   2. Add the field on Gecko `in the environment object <https://searchfox.org/mozilla-central/rev/dd8b5213e4e7760b5fe5743fbc313398b85f8a14/toolkit/components/utils/ClientEnvironment.sys.mjs#31>`_ (even as constant or empty value)
+   3. Add the field on Application-Services RS component `in the RemoteSettingsContext struct <https://github.com/mozilla/application-services/blob/43aa6da9690b2f52d1b3e6255ab2d698f46f47a8/components/remote_settings/src/lib.rs#L43-L78>`_
+   4. Add the field on the Android app `in the RemoteSettingsAppContext <https://searchfox.org/mozilla-central/rev/dd8b5213e4e7760b5fe5743fbc313398b85f8a14/mobile/android/android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsService.kt#47-69>`_
+   5. Add the field on the iOS app `in the RemoteSettingsContext object <https://github.com/mozilla-mobile/firefox-ios/blob/3a2cbe040acb999c6f1589d128f1cfc749e993e5/firefox-ios/Providers/Profile.swift#L782-L799>`_
```
bendk, line 175: I like the links to the Android/ios repos, but I'm slightly worried they might get stale.  Maybe you could link to a searchfox/github search.

bendk (approved):
Looks good to me.  I had a minor suggestion about how to link to firefox-android and firefox-ios.  However you want to deal with that is fine with me.

docs/target-filters.rst
leplatrem: I know...
I pointed a specific commit intentionally, hoping readers would follow the tree if needed 🤷 

Unless I misunderstood your suggestion, I almost have the same concern: if the struct gets renamed, the search won't return anything :/

bendk: > I pointed a specific commit intentionally, hoping readers would follow the tree if needed 🤷

Oh, I missed that.  It makes sense and it seems as good as any other method.

PR #812, opened 2025-03-21
Fix #406: consistency of python versions in CI and Dockerfile

### LINKED ISSUE #406: Check for consistency of various Python versions
We have a few Python versions defined throughout the repo:
```
> rg "3\.11" -g '!*.rst' -g '!.git/*' --hidden

.readthedocs.yaml
6:    python: "3.11"

.circleci/config.yml
236:      - image: cimg/python:3.11.2
252:      - image: cimg/python:3.11.2
263:      - image: cimg/python:3.11.2

poetry.lock
3378:python-versions = "^3.11"

pyproject.toml
50:python = "^3.11"

RemoteSettings.Dockerfile
1:FROM python:3.11.3 as compile
30:FROM python:3.11.3-slim as production

IntegrationTests.Dockerfile
1:FROM python:3.11.3-slim as build
20:FROM python:3.11.3-slim
```

- some versions are pinned to the minor version, and some are pinned to the patch
- sometimes the versions drift. In the output above, we can see that the `Dockerfile` Python versions are at 3.11.3, while the `cimg` tags are at `3.11.2`. At the time of writing this issue, there is no `cimg/python:3.11.3` tag

We should ensure internal consistency for these various Python versions, if possible. We should also alert ourselves when things drift out of sync.

leplatrem:
Fix #406

In 96e4e2b2b14cb560e1b4cd4df06a3fd83b0a40fb I was verifying all pyproject.toml but I find it dumb, and only targetted CI and Dockefiles

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,7 +28,7 @@ jobs:
           key: poetry-dev-${{ hashFiles('poetry.lock') }}
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.13"
+          python-version: "3.13.2"
       - name: Install dependencies
         run: make install
       - name: Run linting and formatting checks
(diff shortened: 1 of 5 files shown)


alexcottner (approved):
I wonder if there is a way for dependabot to keep this all in sync for us.

Issue #804, opened 2025-03-19
When loading permissions from YAML, do not override groups if no members changed


leplatrem:
<img width="527" alt="Image" src="https://github.com/user-attachments/assets/b5b7de65-e171-4d91-bf1e-7c0515dca164" />

This creates a lot of useless entries in the database.

We could probably clean those by comparing the members field, and deleting all history entries where only `last_modified` differs

PR #816, opened 2025-03-24
Ref #804: add script to list groups history and useless entries that can be deleted

leplatrem:
Ref #804

I thought that it can be useful to keep these sorts of scripts, as they serve as examples and can easily be modified for other purposes.

* DEV: 3342 useless group history entries that can be deleted
* STAGE: 11123
* PROD: 9906

Other resources like collections or buckets are less likely to be affected by https://github.com/Kinto/kinto-wizard/pull/325

### DIFF
--- a/bin/groups-history.py
+++ b/bin/groups-history.py
@@ -1,0 +1,126 @@
+"""
+This script will list all groups and find in their history the entries
+that don't modify any attribute, and can thus be deleted.
+
+See mozilla/remote-settings#804
+"""
+import asyncio
+import difflib
+import itertools
+import json
+import os
+from copy import copy
+
+import kinto_http
+
+
+ENV = os.getenv("ENV", "prod")
+SERVER_URL = {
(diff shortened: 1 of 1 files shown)


bin/groups-history.py
```diff
@@ -0,0 +1,127 @@
...
+                else:
+                    print(
+                        f"\n{BOLD}{BLUE}=== {user} modified group '{group_id}' on {date} ==={RESET}"
+                    )
+                    diff_dicts(prev_data, current_data)
+            prev_data = current_data
+
+    if to_delete:
+        print(len(to_delete), "history entries can be deleted")
+        sql = f"UPDATE objects SET delete=true WHERE resource_name = 'history' AND id IN ({','.join(f"'{uuid}'" for uuid in to_delete)});"
```
alexcottner, line 122: nit: We could definitely do this with just a SQL statement, but let's get this history cleaned up.

PR #820, opened 2025-03-25
Fix #793: notify devs when build/publish fails

### LINKED ISSUE #793: Build failure on `main` could notify devs
Image

For example the error on main https://github.com/mozilla/remote-settings/commit/f4e000b9441da6beb0a8a80754fa4c23f3c87d2b was not reported

leplatrem:
Fix #793

Should we notify on success? I'd say we don't since we have Argo doing it already.

* [ ] Figure out if we're allowed to use `mozilla-it/deploy-actions/slack`
* [ ] Set `env.SLACK_CHANNEL_ID_DEVELOPER` in the `build` environment
* [ ] Set `env.SLACK_BOT_TOKEN_BUGID_1796141_20221019` in the `build` environment


leplatrem:
This PR is blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1796141#c10

PR #840, opened 2025-04-01
Updating changes config in sample files, test, and read-me to reflect kinto behavior

alexcottner:
Updating the `changes.resources` config in our tests, sample config, and read-me as this is a "starts with" match on URI's now.

### DIFF
--- a/kinto-remote-settings/tests/changes/test_changes.py
+++ b/kinto-remote-settings/tests/changes/test_changes.py
@@ -169,7 +169,7 @@ class UpdateChangesTest(BaseWebTest, unittest.TestCase):
             "description": "Track modifications of records in Kinto and store "
             "the collection timestamps into a specific bucket "
             "and collection.",
-            "collections": ["/buckets/blocklists"],
+            "collections": ["/buckets/blocklists/"],
             "url": "http://kinto.readthedocs.io/en/latest/tutorials/"
             "synchronisation.html#polling-for-remote-changes",
             "version": __version__,
--- a/config/local.ini
+++ b/config/local.ini
@@ -88,12 +88,12 @@ mail.debug_mailer = true
 #
 
 kinto.changes.resources =
-    /buckets/main
-    /buckets/main-preview
-    /buckets/security-state
-    /buckets/security-state-preview
-    /buckets/blocklists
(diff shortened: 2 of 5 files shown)


leplatrem:
So we are fixing this because, when we specify:
```ini
kinto.changes.resources = /buckets/main
```
It will implicity include `/buckets/main-workspace` and `/buckets/main-preview` too.

Could you elaborate why this behaviour change gives us flexibility?
Like an example of use-case you thought of?
My 2 cents is that if we want to do pattern matching, we would do it with an explicit syntax in config, for example like this:
```ini
kinto.changes.resources = /buckets/main*
```

We could have changed these lines to preserve the previous behavior:

https://github.com/mozilla/remote-settings/blob/f923084babf0abfb217478484ff36be8a781e107/kinto-remote-settings/src/kinto_remote_settings/changes/utils.py#L40-L42

to just:

```py
 matches_included = any( 
     parent_id.startswith(f"{uri}/") for uri in included_resources_uri 
 ) 
```

kinto-remote-settings/tests/changes/test_changes.py
```diff
@@ -169,7 +169,7 @@ def test_changes_capability_exposed(self):
             "description": "Track modifications of records in Kinto and store "
             "the collection timestamps into a specific bucket "
             "and collection.",
-            "collections": ["/buckets/blocklists"],
```
leplatrem, line 172: We should add a dedicated test to specify our behaviour with trailing slashes

alexcottner:
> Could you elaborate why this behaviour change gives us flexibility?
> Like an example of use-case you thought of?

My main idea is to allow more flexibility in buckets. For remote-settings, we have `main`, `main-preview` and `main-workspace`.

But this flexibility leaves the door for us (or other kinto projects) to break collections into more buckets. Like:
`prod-nimbus` and `prod-security`. Then monitor all of them with `/buckets/prod-`.

It felt like a happy accident kinda thing.

leplatrem:
I bypassed rules and merged the fix in the remote settings code to prevent clients to receive push notifications everytime there's a change in `main-workspace`.
We can use this PR to implement smarter matchs if we find it useful, but with unit tests that make behaviour explicit

PR #843, opened 2025-04-02
Fix #842: prevent automatically deleted collections to show up in monitor/changes

### LINKED ISSUE #842: Invalid collection entries in monitor/changes
With the latest code from #837 a new entry with a trailing `,` has showed up in the monitor/changes list:

Image

which breaks most of our checks:

Image

leplatrem:
Fix #842 

* [x] Reproduce issue
* [x] Run "purge" instead of just deleting from storage
* [x] Requires memory backend fix from https://github.com/Kinto/kinto/pull/3520


alexcottner (dismissed):

alexcottner (approved):
should be good now

Issue #836, opened 2025-04-01
Manage history entries of update-heavy ingestion scripts


leplatrem:
```
db=> SELECT data->>'collection_id', COUNT(*) AS count FROM objects WHERE resource_name = 'history' AND data->>'resource_name' = 'collection' GROUP BY data->>'collection_id' ORDER BY 2 DESC LIMIT 20;
           ?column?            | count
-------------------------------+-------
 cert-revocations              | 57047
 intermediates                 | 40130
 nimbus-preview                | 29967
 tracking-protection-lists     | 27084
 crash-reports-ondemand        | 21869
 normandy-recipes-capabilities | 19514
 nimbus-desktop-experiments    | 13114
 normandy-recipes              | 10522
 quicksuggest                  |  7828
```

See https://github.com/Kinto/kinto/pull/3515
See https://github.com/mozilla/remote-settings-ondemand-crashes/issues/5

PR #845, opened 2025-04-03
 Ref #836: Add core cronjob to purge old history entries

leplatrem:
Ref #836

This would be complementary to https://github.com/mozilla-it/webservices-infra/pull/4667

We could run this once a week...

### DIFF
--- a/cronjobs/src/commands/purge_history.py
+++ b/cronjobs/src/commands/purge_history.py
@@ -1,0 +1,36 @@
+from datetime import datetime, timedelta, timezone
+
+from decouple import Csv, config
+
+from . import KintoClient as Client
+
+
+DEFAULT_USER_IDS = (
+    "account:admin, account:cloudservices_kinto_prod, plugin:remote-settings"
+)
+DEFAULT_BUCKETS = "staging, security-state-staging, main-workspace"
+
+
+def utcnow():
+    return datetime.now(timezone.utc)
+
+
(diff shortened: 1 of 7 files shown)


cronjobs/src/commands/purge_history.py
```diff
@@ -0,0 +1,36 @@
...
+    buckets = config("BUCKETS", default=DEFAULT_BUCKETS, cast=Csv())
+
+    limit_date = utcnow() - timedelta(days=max_age_days)
+    limit_timestamp = f'"{int(limit_date.timestamp() * 1000)}"'
+
+    client = Client(server_url=server_url, auth=auth)
+
+    for bucket in buckets:
+        for user_id in user_ids:
+            deleted = client.purge_history(
```
alexcottner, line 33: Looking at the implementation, I wish we had set this up to accept a list of buckets and users rather than calling it one at a time.

alexcottner (approved):

cronjobs/src/commands/purge_history.py
leplatrem: Technically we can pass a list of users:
https://docs.kinto-storage.org/en/latest/api/1.x/filtering.html#multiple-values
We could have passed `in_user_id=",".join(user_ids)`

For several buckets at once we could use a batch request.

But the performance of such a cronjob is not key. Plus, this is a very cheap way to do it in bulks instead of doing one big operation.

PR #846, opened 2025-04-03
Ref #939: ADR about growing objects table

leplatrem:
Ref #939


docs/adr/adr_005_database_objects.md
```diff
@@ -0,0 +1,152 @@
...
+
+In this case, the “parent“ `objects` table acts more like a “routing layer” for the partitions. It doesn't store any data itself.
+
+Since most (all?) queries in Kinto include a `resource_name` filter, using partitions would immediately bring:
+
+- better query performance (smaller sub-tables, parallel execution, ...)
+- better write performance
+- faster indexing
+
+The main downside is that objects triggers and indexes would have to be re-created on each partition table, making future database migrations more complex.
```
alexcottner, line 147: > The main downside is that objects triggers and indexes would have to be re-created on each partition table, making future database migrations more complex.

I don't think this is true for postgres (maybe this was true for older versions though?). I just ran a test with a postgres 15 container (see test script below) and I didn't have to explicitly create indexes or triggers on my partitions.

```sql
-- test setup
create table partition_test (
	part varchar,
	id uuid,
	data jsonb,
	primary key(part, id)
) partition by list(part);

create table partition_test_default partition of partition_test default;
create table partition_test_foo partition of partition_test for values in ('foo');

create table partition_test_log(
	id uuid primary key,
	timestamp timestamp,
	action varchar,
	part varchar,
	data jsonb
);

create function partition_test_logger() returns trigger as $$
	begin
		insert into partition_test_log(id, timestamp, action, part, data)
		select gen_random_uuid(), now(), TG_OP, coalesce(NEW.part, OLD.part), coalesce(NEW.data, OLD.data);
		return NEW;
	end;
$$ language plpgsql;

create trigger tgr_partition_test_modify
	after insert or update on partition_test
	for each row execute function partition_test_logger();

create index idx_partition_test_data on partition_test using gin(data);

-- test the trigger on the primary table applies to partitions
insert into partition_test(part, id, data)
values('a', gen_random_uuid(), '{"a": "is a letter"}'),
('foo', gen_random_uuid(), '{"a": "is a letter"}');

update partition_test set data = '{"b": "is also a letter"}';

select * from partition_test_log;
-- shows audit logs from trigger

-- generate enough test data to ensure an index hit
insert into partition_test(part, id, data)
select f, gen_random_uuid(), ('{"' || prop || '": "is a test"}')::jsonb
from generate_series(1,10000) s
cross join unnest(array['foo', 'bar']) f
cross join unnest(array['a', 'b', 'c', 'd', 'e', 'f']) prop;

-- verify that our queries hit the indexes
explain analyze
select id
from partition_test
where part='foo' and data ? 'foo';
-- shows an index hit on our foo partition table
```

docs/adr/adr_005_database_objects.md
```diff
@@ -0,0 +1,152 @@
...
+
+Also, with the current code architecture, `kinto.core` is not “aware“ of possible values for `resource_name`. Creating partitions for `history`, `bucket`, etc. would either break this separation of concern, and would have to be done elsewhere, like in the `kinto-remote-settings` plugin (for which we currently don't have notions of database migrations).
+
+Note that since our storage implementation lives in *Kinto*, we would impose this partitioning to all users, even for small use-cases.
```
alexcottner, line 151: We could create only the default partition in kinto and allow implementations to handle this themselves. This way we aren't creating opinions for anybody.

docs/adr/adr_005_database_objects.md
```diff
@@ -0,0 +1,152 @@
...
+PostgreSQL would copy data between the source database and this new one using a publisher/subscriber model, replicating data changes (INSERT/UPDATE/DELETE) at the table level, and we could select only the objects used by readers (buckets, collections, and records).
+
+Objects used by the writer, like accounts, history, attachments links, etc. would not be replicated.
+
+We would have some flexibility to control the final dataset (eg. do not include tombstones that are more than 2 years old).
+
+Since performance on the writer is not critical, we could adjust the database resources according to their respective load.
+Any CPU spike on the reader database would not affect user experience.
+
+- **Complexity**: Low. This setup a classic read/write split architecture
```
alexcottner, line 172: It doesn't look like GCP CloudSQL supports partial replication natively, but we could set it up manually with a secondary instance acting as the reader. Offloading our main reads to a secondary node would also allow us to reduce our writer size.

docs/adr/adr_005_database_objects.md
```diff
@@ -0,0 +1,175 @@
...
+ collection    |    346
+ group         |    236
+ bucket        |     26
+(6 rows)
+```
+
+In [mozilla/remote-settings#836](https://github.com/mozilla/remote-settings/issues/836) we introduced new routines that will
+keep the history size under control.
+
+However, still a good proportion of objects (80+% for records) are tombstones that we can't delete from the database since clients rely on them for synchronization.
```
alexcottner, line 59: I wonder if we could remove older tombstone records. Like, do we really care that record A was deleted 11 months ago? If a client pulls a set of changes they'll pull all current state, right?

docs/adr/adr_005_database_objects.md
alexcottner: Expanding on this: Adding a new partition for the table with existing data in it would look something like this (using example tables above).
```sql
alter table partition_test disable trigger tgr_partition_test_modify;

create table partition_test_a (like partition_test including constraints);
with moving as (
	delete from partition_test where part = 'a returning *
)
insert into partition_test_a 
select * from moving;

alter table partition_test attach partition partition_test_a for values in ('a');

alter table partition_test enable trigger tgr_partition_test_modify;
```

alexcottner (approved):

PR #931, opened 2025-06-16
fix - fixing doc warnings for highlighting failures on grafana sql blocks

alexcottner:
We are getting a warning on docs build for sql code blocks that have the `$__timeFilter` grafana parameter.
Removing the sql type from the code blocks.


leplatrem (approved):
👌

Issue #897, opened 2025-05-22
On release, tag the existing image of main instead of rebuilding


leplatrem:
See https://github.com/mozilla-services/autograph/pull/1037

Downside: if release is tagged too quickly after merge on main, the image is not yet available and fails

alexcottner:
Implemented! This takes publishing a new release from 10 minutes to less than 2 minutes.

PR #937, opened 2025-06-25
Ref #897 - Adjust server release process to only re-tag on release rather than rebuild

alexcottner:
Ref #897 - re-tagging the existing server image on release rather than rebuilding.
Core cronjob and browser tests images only use `latest` tags, adjusting so they aren't rebuilt on release.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -112,22 +112,6 @@ jobs:
           platforms: linux/amd64,linux/arm64
           cache-from: type=gha # Load cache from GitHub Actions
           cache-to: type=gha,mode=max # Save cache to GitHub Actions
-      - name: Set up crane # copied from https://github.com/imjasonh/setup-crane/blob/main/action.yml
-        shell: bash
-        run: |
-          set -ex
-
-          out=crane
-          os=${{ runner.os }}
-          tag=$(curl -s -u "username:${{ github.token }}" https://api.github.com/repos/google/go-containerregistry/releases/latest | jq -r '.tag_name')
-          arch=$(uname -m)
-          tmp=$(mktemp -d)
-          cd "${tmp}"
-          curl -fsL "https://github.com/google/go-containerregistry/releases/download/${tag}/go-containerregistry_${os}_${arch}.tar.gz" | tar xz "${out}"
-          chmod +x "${tmp}/${out}"
-          PATH=${PATH}:${tmp}
-          echo "${tmp}" >> "${GITHUB_PATH}"
-          echo "${{ github.token }}" | crane auth login ghcr.io --username "dummy" --password-stdin
       - name: Copy from Google Artifact Registry to Docker Hub
         if: github.event_name != 'pull_request'
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -83,18 +98,60 @@ jobs:
...
+          TAGS: |
+            ${{ steps.metahub.outputs.tags }}
+          SRC: ${{ env.GAR_LOCATION }}-docker.pkg.dev/${{ env.GCP_PROJECT_ID }}/${{ env.GAR_REPOSITORY}}/autograph:sha-${{ github.sha }}
+        run: |
+          crane digest "${SRC}"
+          crane manifest "${SRC}"
+          for tag in $TAGS; do
+            crane copy "${SRC}" "${tag}"
+          done
+      - name: Copy image from Google Artifact Registory to Google Artifact Registory as semver tag
```
say-yawn, line 143:
```suggestion
      - name: Copy image from Google Artifact Registry to Google Artifact Registry as semver tag
```
nit

say-yawn (changes requested):
First attempt. Is there a need to keep the `pull_request` as an event trigger?

alexcottner:
> First attempt. Is there a need to keep the `pull_request` as an event trigger?

We wanna keep the `pull_request` event so the containers are built when the PR checks run. But we won't push them to dockerhub or GAR.

.github/workflows/publish.yaml
```diff
@@ -83,18 +98,60 @@ jobs:
...
           target: production  # See multi-stage build
           push: ${{ github.event_name != 'pull_request' }}
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          tags: ${{ steps.metagar.outputs.tags }}
+          labels: ${{ steps.metagar.outputs.labels }}
           platforms: linux/amd64,linux/arm64
           cache-from: type=gha # Load cache from GitHub Actions
           cache-to: type=gha,mode=max # Save cache to GitHub Actions
+      - name: Set up crane # copied from https://github.com/imjasonh/setup-crane/blob/main/action.yml
```
smarnach, line 115: Why do we need crane? ~Retagging images should be [possible with the normal Docker CLI](https://docs.docker.com/reference/cli/docker/buildx/imagetools/create/), which is already installed on the runner image.

.github/workflows/publish.yaml
```diff
@@ -50,17 +51,31 @@ jobs:
...
-            ${{ env.DOCKERHUB_IMAGE_NAME }}
             ${{ env.GAR_LOCATION }}-docker.pkg.dev/${{ env.GCP_PROJECT_ID }}/${{ env.GAR_REPOSITORY }}/${{ env.GAR_IMAGE_NAME }}
-          # https://github.com/marketplace/actions/docker-metadata-action#tags-input
           tags: |
-            type=raw,value=latest
-            type=raw,value=${{ env.LATEST_TAG }}
+            type=raw,value=${{ env.LATEST_TAG }},enable=${{ github.event_name == 'push' }}
+            type=sha,format=long,enable=${{ github.event_name == 'push' }}
+            type=semver,pattern={{raw}},enable=${{ github.event_name == 'release' }}
+      - name: Docker Metadata for Docker Hub
```
smarnach, line 67: Do we still need to push to Docker Hub? For crash ingestion, we are only using GAR now. It's still possible to make the GAR images public if desired.

PR #962, opened 2025-07-30
Add signer user id to capabilities

leplatrem:


alexcottner (approved):
Need to update one unit test check but looks good

PR #978, opened 2025-08-05
Fix #977 - Returning 400 responses to clients when `_expected` value is less than `_since` value

### LINKED ISSUE #977: Handle clients sending stale expected values
Occasionally, we see clients sending stale `_expected` values vs their `_since` timestamps. We should be able to reject or redirect these similar to stale `_since` values.

alexcottner:
Ref #977 - Rejecting requests with `_expected` values lower than `_since` values. This will prevent a rare problem with them trying to go back in time.

- [x] Add logic to check `_expected` against `_since` and return a 400 response
- [x] Add unit tests


alexcottner:
> I am trying to understand how this differs from the behaviour where we don't do anything on the server. But I don't find any major improvement.

I think this is correct. I was hoping we could do a helpful redirect, but after talking further I think the right answer is to just return a 400 for these requests and the clients will catch up shortly after anyway.

leplatrem (approved):

PR #982, opened 2025-08-07
Switching stale 400 response to 204 so gcp cdn can cache it

alexcottner:
Ref #977 - Switching response code from 400 to 204 so gcp cdn can cache correctly


leplatrem (approved):
👌

PR #983, opened 2025-08-07
Ref #977 - Moving back to 400 responses for stale expected values after fastly migration

alexcottner:
Ref #977 - Moving back to 400 responses for stale expected values after fastly migration

Leaving in draft for now, will merge when we are ready to test.

### DIFF
--- a/kinto-remote-settings/src/kinto_remote_settings/changes/views.py
+++ b/kinto-remote-settings/src/kinto_remote_settings/changes/views.py
@@ -186,7 +186,7 @@ def _handle_stale_expected(request):
 
     # Check if client is trying to go back in time, return 400 response
     if qs_expected > 0 and qs_expected < qs_since:
-        response = httpexceptions.HTTPNoContent()
+        response = httpexceptions.HTTPBadRequest()
         cache_seconds = int(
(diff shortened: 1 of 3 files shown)


kinto-remote-settings/src/kinto_remote_settings/changes/views.py
```diff
@@ -186,7 +186,7 @@ def _handle_stale_expected(request):
 
     # Check if client is trying to go back in time, return 400 response
     if qs_expected > 0 and qs_expected < qs_since:
-        response = httpexceptions.HTTPNoContent()
+        response = httpexceptions.HTTPBadRequest()
```
leplatrem, line 189: See changes of #982 
```suggestion
        raise_invalid(
            request,
            name="_expected",
            location="querystring",
            description="`_expected` parameter cannot be lower than `_since`",
```

leplatrem (changes requested):

leplatrem:
Please note that 400 requests won't be cached by the CDN (see https://github.com/mozilla/webservices-infra/pull/7525)

(This should be OK though since we don't reach the DB before returning them)

PR #1060, opened 2025-11-04
cronjob: expire orphan attachments, set `custom_time` instead of releasing hold

leplatrem:
We changed our strategy in https://github.com/mozilla/webservices-infra/pull/8059

We now rely on custom_time to purge old orphan attachments

### DIFF
--- a/cronjobs/src/commands/expire_orphan_attachments.py
+++ b/cronjobs/src/commands/expire_orphan_attachments.py
@@ -19,7 +19,7 @@ STORAGE_BUCKET_NAME = os.getenv(
 
 def expire_orphan_attachments(event, context):
     """
-    This cronjob will set the current time on orphan attachments.
+    This cronjob will set the custom time field on orphaned attachments to the current time.
     We then have a retention policy on GCS bucket that will
     soft-delete these objects after N days.
 


cronjobs/src/commands/expire_orphan_attachments.py
```diff
@@ -18,10 +19,9 @@
 
 def expire_orphan_attachments(event, context):
     """
-    This cronjob will release holds on orphan attachments.
-    See https://cloud.google.com/storage/docs/holding-objects
-    And then GCS will start the count down, and when the retention
-    period is over, it will soft-delete them.
+    This cronjob will set the current time on orphan attachments.
```
smarnach: Maybe "This cronjob will set the custom time field on orphaned attachments to the current time."?

cronjobs/src/commands/expire_orphan_attachments.py
```diff
@@ -71,5 +71,5 @@ def expire_orphan_attachments(event, context):
             print(
                 f"Marking orphan attachment gs://{STORAGE_BUCKET_NAME}/{blob.name} for deletion"
             )
-            blob.event_based_hold = False
+            blob.custom_time = datetime.now(timezone.utc)
```
smarnach: I presume you dobule-checked this actually takes a `datetime` object rather than, say, a Unix timestamp.

smarnach (approved):

cronjobs/src/commands/expire_orphan_attachments.py
leplatrem: yes!

PR #1071, opened 2025-11-10
RMST-235: Add metrics to git-reader

leplatrem:
* [x] Add metrics
* [x] Add tests
* [x] Adjust tests and settings reading

### DIFF
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -28,7 +28,7 @@ HERE = pathlib.Path(__file__).parent.resolve()
 VERSION = "0.0.1"
 # The API is served under /v2 prefix since /records endpoints present in /v1
 # are not implemented.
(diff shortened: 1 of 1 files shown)


git-reader/app.py
```diff
@@ -411,7 +474,31 @@ async def lifespan(app: FastAPI):
...
+) -> Response:
+    global METRICS
+
+    start_time = time.time()
+    response = await call_next(request)
+    elapsed_sec = time.time() - start_time
+
+    labels = (
+        request.method,
+        request.scope["route"].name if "route" in request.scope else "unknown",
```
alexcottner: Making sure I understand what `request.scope["route"].name` will be: This will be things like `attachments` and `broadcasts` and `cert_chain`, right?

PR #1075, opened 2025-11-12
git-export: Delete attachments once every 2 days by default

leplatrem:
Duplicating the git-export cronjob definition is a PIT*   (duplicating volumes, env vars, secrets, ...). Just to set this flag. Plus having 2 cronjobs can run concurrent operations, not great.

This would be an alternative approach. Not the most elegant, but I think it would do it since we run the cronjob every 10min.

Thoughts?

### DIFF
--- a/cronjobs/src/commands/git_export.py
+++ b/cronjobs/src/commands/git_export.py
@@ -74,8 +74,8 @@ MIN_TAGS_PER_COLLECTION_COUNT = config(
 # This avoids running this potentially expensive operation on every cronjob run.
 # And to avoid duplicating the cronjob definition twice just to set this env var.
 _now = datetime.datetime.now(datetime.timezone.utc)
-_IS_EVEN_DAY = _now.weekday() % 2 == 0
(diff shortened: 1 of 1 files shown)


cronjobs/src/commands/git_export.py
```diff
@@ -68,8 +69,16 @@
 MIN_TAGS_PER_COLLECTION_COUNT = config(
     "MIN_TAGS_PER_COLLECTION_COUNT", default=2, cast=int
 )
+
+# By default, we delete unreachable attachments only once every 2days between 12:00 and 12:15 UTC
+# This avoids running this potentially expensive operation on every cronjob run.
+# And to avoid duplicating the cronjob definition twice just to set this env var.
+_now = datetime.datetime.now(datetime.timezone.utc)
+_IS_EVEN_DAY = _now.day % 2 == 0
```
alexcottner: Obligatory note: this will cause 3-day gaps 7 times a year (sometimes 8) because calendaring is hard.

alexcottner (approved):

PR #1119, opened 2025-12-30
RMST-260: git-export: update metadata even if records haven't changed

leplatrem:
https://mozilla-hub.atlassian.net/browse/RMST-260


alexcottner (approved):
Have a suggestion but we could do that as a follow-up

PR #1243, opened 2026-04-10
RMST-373 - Returning welcome and broadcasts responses when git folder is empty

alexcottner:
[RMST-373](https://mozilla-hub.atlassian.net/browse/RMST-373)
When git-reader's data has not been fully initialized, we don't return great responses on the root or broadcast endpoints. This aims to improve that.

- Broadcast endpoint now returns an empty object, rather than just 500.
    - Note: actual http response code is still 200 to not break existing logic
- Root response now returns a 503 with an error message

Unsure if we can improve anything beyond this without the git state.

### DIFF
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -81,7 +81,9 @@ METRICS = {
         buckets=[0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1.0, float("inf")],
     ),
 }
-NO_GIT_ERROR = "Unable to load state from git. Has the update job completed?"
+NO_GIT_ERROR = (
+    "Unable to load state from `GIT_REPO_PATH`. Has the `gitupdate` job completed?"
+)
 
 # Augment the default mimetypes with our own.
 mimetypes.init(files=[HERE / "mimetypes.txt"])
@@ -162,7 +164,7 @@ def get_settings() -> Settings:
 def get_last_modified(settings: Settings = Depends(get_settings)) -> int | float:
     try:
         return os.path.getmtime(settings.git_repo_path)
-    except Exception:
+    except OSError:
         return -1
 
 
@@ -217,11 +219,6 @@ class HelloResponse(BaseModel):
 class BroadcastsResponse(BaseModel):
     broadcasts: dict[str, str] = Field(description="Broadcasts")
     code: int = Field(description="HTTP status code")
-    detail: str | None = Field(
-        None,
-        description="Error message detailing the potential problem",
-        exclude_if=lambda v: v is None,
-    )
 
 
 class ChangesetResponse(BaseModel):
@@ -465,9 +462,8 @@ class GitService:
         except Exception as e:
             logger.error(e)
             return {
-                "broadcasts": {"remote-settings/monitor_changes": '"-1"'},
-                "code": 503,
-                "detail": NO_GIT_ERROR,
+                "broadcasts": {},
+                "code": 200,
             }
 
     def get_cert_chain(self, pem: str) -> str:
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -81,6 +81,7 @@ METRICS = {
         buckets=[0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1.0, float("inf")],
     ),
 }
+NO_GIT_ERROR = "Unable to load state from git. Has the update job completed?"
 
 # Augment the default mimetypes with our own.
 mimetypes.init(files=[HERE / "mimetypes.txt"])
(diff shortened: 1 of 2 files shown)


git-reader/app.py
```diff
@@ -441,9 +458,17 @@ def get_broadcasts(self):
...
-        content = json.loads(bcontent.decode("utf-8"))
-        return content
+        try:
+            bcontent = self._get_file_content("broadcasts.json")
+            content = json.loads(bcontent.decode("utf-8"))
+            return content
+        except Exception as e:
+            logger.error(e)
+            return {
+                "broadcasts": {"remote-settings/monitor_changes": '"-1"'},
```
leplatrem: Why `-1`? 
Since the push server considers this string as opaque, won't it send a push notification with `-1` in it? 
I would return an empty `broadcasts: {}` object instead.

git-reader/app.py
```diff
@@ -81,6 +81,7 @@
         buckets=[0.001, 0.005, 0.01, 0.05, 0.1, 0.5, 1.0, float("inf")],
     ),
 }
+NO_GIT_ERROR = "Unable to load state from git. Has the update job completed?"
```
leplatrem:
```suggestion
NO_GIT_ERROR = "Unable to load state from `GIT_REPO_PATH`. Has the `gitupdate` job completed?"
```

git-reader/app.py
```diff
@@ -441,9 +458,17 @@ def get_broadcasts(self):
...
+        try:
+            bcontent = self._get_file_content("broadcasts.json")
+            content = json.loads(bcontent.decode("utf-8"))
+            return content
+        except Exception as e:
+            logger.error(e)
+            return {
+                "broadcasts": {"remote-settings/monitor_changes": '"-1"'},
+                "code": 503,
+                "detail": NO_GIT_ERROR,
```
leplatrem: Do we really need this here?

git-reader/app.py
```diff
@@ -205,6 +217,11 @@ class HelloResponse(BaseModel):
 class BroadcastsResponse(BaseModel):
     broadcasts: dict[str, str] = Field(description="Broadcasts")
     code: int = Field(description="HTTP status code")
+    detail: str | None = Field(
+        None,
+        description="Error message detailing the potential problem",
+        exclude_if=lambda v: v is None,
+    )
```
leplatrem: We could skip this `detail` field if it's not consumed by the push server. Usually we use the `/__heartbeat__` endpoint to troubleshoot errors no?

git-reader/app.py
```diff
@@ -158,12 +159,23 @@ def get_settings() -> Settings:
     return settings
 
 
+def get_last_modified(settings: Settings = Depends(get_settings)) -> int | float:
+    try:
+        return os.path.getmtime(settings.git_repo_path)
+    except Exception:
```
leplatrem: Would it bring us anything to only catch OSError ?

git-reader/app.py
```diff
@@ -158,12 +159,23 @@ def get_settings() -> Settings:
...
+        return -1
+
+
 @lru_cache(maxsize=1)
-def get_repo(settings: Settings = Depends(get_settings)) -> pygit2.Repository:
+def get_repo(
+    settings: Settings = Depends(get_settings),
+    cache_bust: int = Depends(get_last_modified),
```
leplatrem, line 174: oh is that enough to invalidate the `@lru_cache` decorator?

git-reader/tests/test_api.py
```diff
@@ -344,6 +349,15 @@ def test_hello_view(api_client):
     assert resp.headers["cache-control"] == "max-age=3600"
 
 
+def test_hello_view_git_error(app, api_client):
+    app.dependency_overrides[get_repo] = empty_get_repo
```
leplatrem, line 358: I believe that this test does not exactly test the behaviour of the cached `get_repo()` dependency in the app code. 
Especially the behaviour of the `@lru_cache` when the folder starts with being empty or missing, and then is available

git-reader/app.py
alexcottner: If I'm understanding it correctly, it's just a memoized cache. So if any of the params change the result should change. It appears to work this way in local testing.

git-reader/app.py
alexcottner: I was trying to give an indication that something wasn't quite right. But returning an empty object should do that too.

git-reader/app.py
leplatrem: you are completely right 👌  lru-cache hashes the parameter for the cache key 💯

git-reader/app.py
alexcottner: The heartbeat and root paths will return an error, so that should be good enough.

git-reader/app.py
alexcottner: I'll adjust it to be more focused. But in this case my thinking was no matter what exception is being thrown, it means we can't read the git repo (yet).

git-reader/tests/test_api.py
alexcottner: It does not test the lru_cache logic. I'll add some tests for that scenario. Although I don't tend to add unit tests to validate dependencies, this might be good to ensure it behaves as expected and doesn't change.

alexcottner: Added test

leplatrem (approved):

PR #1266, opened 2026-05-04
RMST-388: Weekly cronjob to report consumption movers on Slack

leplatrem:
The idea is to send a weekly post on Slack that will give a summary of top cost increase and decrease

```
*Bandwidth Consumption*
Period: Apr 26 → May 3

*Top increases* (Compared against previous 90-day daily average.)
- `crash-reports-ondemand`: +629.1%
- `startup`: +12.1%
- `nimbus-desktop-experiments`: +8.6%

*Top decreases*
- `quicksuggest-amp`: -91.2%
- `cert-revocations`: -53.6%
- `intermediates`: -36.6%
- `newtab-wallpapers-v2`: -63.2%
- `translations-models-v2`: -25.7%
```

### DIFF
--- a/cronjobs/src/commands/consumption_movers.py
+++ b/cronjobs/src/commands/consumption_movers.py
@@ -1,3 +1,4 @@
+import heapq
 from datetime import date, timedelta
 
 import requests
@@ -35,15 +36,6 @@ GROUP BY collection_id
 """
 
 
-def execute_query(client: bigquery.Client, query: str) -> list[dict]:
-    rows = client.query(query).result()
-    return [dict(row) for row in rows]
-
-
-def rows_to_dict(rows: list[dict], idxfield: str, valuefield: str) -> dict:
-    return {row[idxfield]: row[valuefield] for row in rows}
-
-
 def consumption_movers() -> None:
     """
     Identify the collections with the biggest increases and decreases in bandwidth consumption,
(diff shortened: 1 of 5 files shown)


cronjobs/src/commands/consumption_movers.py
```diff
@@ -0,0 +1,130 @@
...
+    last_rows = execute_query(
+        client,
+        COLLECTIONS_AVERAGE.format(
+            interval_days_start=LAST_PERIOD_DAYS,
+            interval_days_end=0,
+        ),
+    )
+
+    previous = rows_to_dict(previous_rows, "collection_id", "avg_size")
+    last = rows_to_dict(last_rows, "collection_id", "avg_size")
```
smarnach: I think you can achieve the same result using

```
    previous = dict(client.query(previous_query).result())
    last = dict(client.query(last_query).result())
```

The functions `execute_query()` and `rows_to_dict()` aren't really needed.

(This assumes that the query returns exactly two columns, with the first column representing keys and the second column representing values.)

(Feel free to ignore, as usual.)

cronjobs/src/commands/consumption_movers.py
```diff
@@ -0,0 +1,130 @@
...
+            # Skip new collections.
+            continue
+
+        # Skip small collections to avoid noise.
+        if last_avg < MIN_AVG_BYTES and previous[collection_id] < MIN_AVG_BYTES:
+            continue
+
+        previous_avg = previous[collection_id]
+        delta = last_avg - previous_avg
+        pct_change = (last_avg / previous_avg - 1) * 100
```
smarnach: This is a potential division by zero.

cronjobs/src/commands/consumption_movers.py
```diff
@@ -0,0 +1,130 @@
...
+        movers.append(
+            {
+                "collection_id": collection_id,
+                "pct_change": pct_change,
+                "delta": delta,
+            }
+        )
+
+    top_increase = sorted(
```
smarnach: I suggest using [`heapq.nlargest()`](https://docs.python.org/3/library/heapq.html#heapq.nlargest).

cronjobs/src/commands/consumption_movers.py
```diff
@@ -0,0 +1,130 @@
...
+        [m for m in movers if m["pct_change"] > 0],
+        key=lambda m: m["delta"],
+        reverse=True,
+    )[:TOP_N]
+
+    top_decrease = sorted(
```
smarnach: I suggest using [`heapq.nsmallest()`](https://docs.python.org/3/library/heapq.html#heapq.nsmallest).

smarnach (approved):

PR #1269, opened 2026-05-06
Fix #370: do not mount src and config volumes in CI

### LINKED ISSUE #370: Avoid `/app` volume mounts in CI
We typically mount the root directory as a volume in docker-compose to speed up the cycle time of local development. This recently bit us with file permissions issues in #368.

To avoid this, we should find a way to not mount this volume in testing environments.

leplatrem:
https://github.com/mozilla/remote-settings/issues/370

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -120,6 +120,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       DOCKER_CACHE: /tmp/docker-cache
+      CONFIG_MOUNT: ""
+      SRC_MOUNT: ""
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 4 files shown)


alexcottner (approved):
til about docker compose override files

leplatrem:
> til about docker compose override files

Do you think we should leave the override file for devs, and add an explicit `-f docker-compose.dev.yml` to our Makefile?

alexcottner:
> > til about docker compose override files
> 
> Do you think we should leave the override file for devs, and add an explicit `-f docker-compose.dev.yml` to our Makefile?

Having an explicit `docker-compose.ci.yml` file that makes these changes just for CI is the most intuitive to me.
Not being familiar with this before, I probably would have looked at having the github workflow do it's own specific thing instead.
But if the makefile is handling it for me then it won't bother me whatever way we go.

Issue #1267, opened 2026-05-05
Introduction of Granian in v42 introduce container startup delay/issue


leplatrem:
an error occurs during startup with docker compose:
```
remote-settings ->        | urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=8888): Max retries exceeded with url: /v1/buckets/main-workspace/collections/quicksuggest (Caused by NewConnectionError("HTTPConnection(host='localhost', port=8888): Failed to establish a new connection: [Errno 111] Connection refused"))
```
/cc @luc-lisi

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/RMST-389)

leplatrem:
Root cause:

- uwsgi binds the listening socket in the master process before the app finishes importing
- Granian binds once the `APP` is loaded

For a few msecs, Granian is importing the python code before it spawns the worker.

We can see with this:
```py
# slow.py
import time

print("importing...")
time.sleep(5)
print("done")

async def app(scope, receive, send):
    assert scope['type'] == 'http'

    await send({
        'type': 'http.response.start',
        'status': 200,
        'headers': [
            [b'content-type', b'text/plain'],
        ],
    })
    await send({
        'type': 'http.response.body',
        'body': b'Hello, world!',
    })
```

```
$ uv run granian --interface asgi slow:app

[INFO] Starting granian (main PID: 39575)
[INFO] Listening at: http://127.0.0.1:8000
[INFO] Spawning worker-1 with PID: 39579
importing...
done
[INFO] Started worker-1
```

And while "importing", connection is refused.
Granian does not seem to have any option/flag to change that.

Short term bandage will be to add a heathcheck in docker compose.

But maybe we could look into our code base to improve load time by using lazy imports (for example Sentry probes installed frameworks at import time)

leplatrem:
@luc-lisi the container now has `curl` installed, so you should be able to add a simple healthcheck in your dockercompose like we did in #1270

Please let us know if that works and reopen if needed

luc-lisi:
Thank you, @leplatrem! We will take a look 🙂

PR #1272, opened 2026-05-06
Ref #1267: compile bytecode during container build

leplatrem:


alexcottner (approved):
what kind of perf difference should we be expecting or looking for?

leplatrem:
> what kind of perf difference should we be expecting or looking for?

There won't be any perf difference at runtime. The container will just start faster. Ratio of benefits vs. cost of the PR is high, even if we are talking about a hundred of ms 😅

PR #1273, opened 2026-05-06
Add project_name, project_version, url to Slack notification context

leplatrem:
This would allow to define the notification as ``✨ collection {collection_id} was created on {settings[project_name]}`` instead of having to duplicate the notification definition on each environment.

See also https://github.com/Kinto/kinto-emailer/pull/510


leplatrem:
~~Oh, no, we have `'[{settings[project_name]}]` in kinto-emailer~~
update: replaced `{project_name} with {settings[project_name]}`

alexcottner (approved):

PR #1297, opened 2026-05-13
Fix #1293: build the linux/arm64 in parallel of the linux/amd64

### LINKED ISSUE #1293: Building linux/arm64 containers is slow
And these images are not used in our deployments.

Can we at least only build them on release for example? 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/RMST-395)

leplatrem:
Fix #1293

(Done with `claude`: "build linux/arm64 and linux/amd64 in parallel")

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -224,7 +224,7 @@ jobs:
         run: |
           tag_args=$(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$METADATA_JSON")
           digest_refs=$(printf "${GAR_IMAGE}@sha256:%s " *)
-          docker buildx imagetools create $tag_args $digest_refs
+          docker buildx imagetools create "$tag_args" "$digest_refs"
       - name: Copy from Google Artifact Registry to Docker Hub
         env:
           TAGS: |
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -24,23 +24,34 @@ env:
 jobs:
   server_container:
     env:
-      DOCKERHUB_IMAGE_NAME: mozilla/remote-settings
       GAR_IMAGE_NAME: remote-settings
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -24,23 +24,34 @@ env:
 jobs:
   server_container:
     env:
-      DOCKERHUB_IMAGE_NAME: mozilla/remote-settings
       GAR_IMAGE_NAME: remote-settings
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
alexcottner, line 29: This can probably be true? If one fails we don't need to keep going.

.github/workflows/publish.yaml
```diff
@@ -128,29 +119,120 @@ jobs:
...
+    if: github.event_name != 'pull_request'
+    needs: server_container
+    env:
+      DOCKERHUB_IMAGE_NAME: mozilla/remote-settings
+      GAR_IMAGE_NAME: remote-settings
+      LATEST_TAG: ${{ needs.server_container.outputs.latest_tag }}
+    runs-on: ubuntu-latest
+    environment: build
+    steps:
+      - name: Download digests
```
alexcottner, line 168: Have you seen this yet? If you have, was there a reason you were avoiding it?
https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners
That feels a little easier to wrap my head around

alexcottner (changes requested):
I _think_ using the `docker/github-builder` build action could make this easier.

.github/workflows/publish.yaml
leplatrem: I don't think it supports auth via Workload Identity Federation, I need to see what workaround
And that our org allows this action too

leplatrem:
Screenshot 2026-05-27 at 17 44 32

PR #1310, opened 2026-05-19
Fix #788: do not raise 400 on multiple changes involving attachments

### LINKED ISSUE #788: Users stuck with `400 Attachment metadata cannot be modified`
Collection is in work-in-progress. 
Users face the same error when trying to request review or rollback changes: `400 Attachment metadata cannot be modified`

Deleting records that are in common between main-workspace and main-preview/main, and then rollbacking worked.

Note: we could reproduce on stage with the search-config-icons:
- rollback the collection
- update the attachment of a record
- request review

Note: could not reproduce on stage with the product-integrity collection 

https://github.com/user-attachments/assets/37aa72d5-8d3e-4ca4-b43d-805aa81606c1

Note: Similar bug existing 7 years ago https://bugzilla.mozilla.org/show_bug.cgi?id=1470812
When copying the records from one bucket to another (ie. in rollback, review request or approval), we set a [special flag](https://github.com/Kinto/kinto-attachment/blob/dea66d776ec5d07c3811f4294e5afed987ed8dc9/src/kinto_attachment/listeners.py#L34) on the request. For some reasons, it does not always get set!

https://github.com/mozilla/remote-settings/blob/d5f24cd32805a6c6436e1f9bf8b3df0a03409a55/kinto-remote-settings/src/kinto_remote_settings/signer/listeners.py#L126-L127

leplatrem:
Fix #788

### DIFF
--- a/kinto-remote-settings/src/kinto_remote_settings/signer/utils.py
+++ b/kinto-remote-settings/src/kinto_remote_settings/signer/utils.py
@@ -238,18 +238,14 @@ def notify_resource_event(
     # When kinto-signer copies record from one place to another,
     # it simulates a resource event. Since kinto-attachment
     # prevents from updating attachment fields, it throws an error.
-    # The following flag will disable the kinto-attachment check.
-    # See https://github.com/Kinto/kinto-signer/issues/256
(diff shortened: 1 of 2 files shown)


alexcottner (approved):
❤️

PR #1341, opened 2026-06-02
Add typos check in CI

leplatrem:
Added `typos` in CI.

Ran it locally with `-w` to fix typos. 

Shall we add it to `make lint` (and thus expect developers to have it installed?)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,15 +25,8 @@ jobs:
         env:
           ACTIONLINT_VERSION: 1.7.7
           ACTIONLINT_SHA256: 023070a287cd8cccd71515fedc843f1985bf96c436b7effaecce67290e7e0757  # pragma: allowlist secret
-      - name: Check typos
-        run: |
-          curl --fail -L "https://github.com/crate-ci/typos/releases/download/v${{ env.TYPOS_VERSION }}/typos-v${{ env.TYPOS_VERSION }}-x86_64-unknown-linux-musl.tar.gz" -o typos.tar.gz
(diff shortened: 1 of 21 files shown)


.github/workflows/ci.yaml
```diff
@@ -25,6 +25,15 @@ jobs:
         env:
           ACTIONLINT_VERSION: 1.7.7
           ACTIONLINT_SHA256: 023070a287cd8cccd71515fedc843f1985bf96c436b7effaecce67290e7e0757  # pragma: allowlist secret
+      - name: Check typos
+        run: |
+          curl --fail -L "https://github.com/crate-ci/typos/releases/download/v${{ env.TYPOS_VERSION }}/typos-v${{ env.TYPOS_VERSION }}-x86_64-unknown-linux-musl.tar.gz" -o typos.tar.gz
```
alexcottner: I think we should have rust/cargo on the ubuntu-latest image already. I think we could make this a little easier like:
```
cargo install typos-cli --version ${{ env.TYPOS_VERSION }} --locked
typos
```

alexcottner: If we don't want a specific version, we could just leave that part empty and it'll pull the latest release.

alexcottner:
> Shall we add it to make lint (and thus expect developers to have it installed?)

Some people (clearly not me :grimacing: ) might already have a spell checker installed in their editor. But I do think we should add this for consistency. Maybe as a new option? `make spell`?

alexcottner (approved):
I think this is a good addition to checks. Have one suggestion.

leplatrem:
Good call

I switched to Cargo, and added it to Makefile as optional

PR #1364, opened 2026-06-17
Making git-reader return 206 for older expected than since values

alexcottner:
Git-reader should return a 206 instead of a 400 when clients request data with an older `_expected` value than the `_since` value.
This will mirror the existing v1 behavior, which is what clients are expecting.

Client-side code ([link](https://searchfox.org/firefox-main/source/services/settings/remote-settings.sys.mjs#462)) in `remote-settings.sys.mjs` is expecting a 400 to mean the etag should be cleared so it can resync. But this is a behavior change from v1.

### DIFF
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -678,7 +678,7 @@ def monitor_changes(
 ):
     if _since and _expected > 0 and _expected < _since:
         raise HTTPException(
-            status_code=400,
+            status_code=206,
             detail="_expected must be superior to _since if both are provided",
         )
 
@@ -713,7 +713,7 @@ def collection_changeset(
 ):
     if _since and _expected > 0 and _expected < _since:
         raise HTTPException(
-            status_code=400,
+            status_code=206,
             detail="_expected must be superior to _since if both are provided",
         )
 
--- a/git-reader/tests/test_api.py
+++ b/git-reader/tests/test_api.py
@@ -417,7 +417,7 @@ def test_monitor_changes_view_filtered_bad_since(api_client):
     resp = api_client.get(
         "/v2/buckets/monitor/collections/changes/changeset?_since=223456789&_expected=123456789"
     )
-    assert resp.status_code == 400
+    assert resp.status_code == 206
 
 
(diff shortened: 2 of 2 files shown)


leplatrem:
> This will mirror the existing v1 behavior,

I can't recall the reason why we did use 206 in the first place. Wasn't it because of GCP caching limitation on 400? 

> This will mirror the existing v1 behavior

`206 Partial Content` seems like a hack. If we can stick with `400` I think it matches more what we want for a future-proof v2

> which is what clients are expecting.

In what sense the client expects 206?

Couldn't we switch v1 to 400 instead?

If we really have no choice, then yeah let's do this :)

alexcottner:
Added a comment to the jira ticket ([link](https://mozilla-hub.atlassian.net/browse/RMST-439?focusedCommentId=1502656)) but the evidence seems pretty strong here.

image
image

alexcottner:
> I can't recall the reason why we did use 206 in the first place. Wasn't it because of GCP caching limitation on 400?

I think the GCP limitation was the big reason before. But I don't think we were returning 400 responses for long enough to encounter this secondary problem.

> In what sense the client expects 206?

I think the main thing here is if we return a 206, the clients don't do anything and keep working with their current data. And will catch up on the next sync. But if we return a 400, the etag value is cleared and they could go backwards.

leplatrem:
> for long enough to encounter this secondary problem.

I would like to understand this secondary problem better, we should document it carefully if what we are solving here makes sense. And add a paragraph here https://remote-settings.readthedocs.io/en/latest/client-specifications.html#endpoints

> But if we return a 400, the etag value is cleared and they could go backwards.

I don't get this part. Why clearing the `?_since` value would make the client go backwards? (Is `_since` really a factor of cache busting? Isn't `_expected`?)
If that is a behaviour we observe, we should try to fix it on the client, no?

I fear that deploying this change on the server will leave us with this hack for many years. It looks like we have the opportunity to start clean with v2, can we make sure this is really what we want? Please help me be convinced 😬

alexcottner:
Closing in favor of client-side changes.

PR #1388, opened 2026-06-24
git-reader - improve collection changeset performance by caching the sorted refs

alexcottner:
[RMST-438](https://mozilla-hub.atlassian.net/browse/RMST-438)
When investigating CPU spikes, found that the `raw_listall_references` call was responsible for 75% of the effort in `get_collection_changeset`.

Caching the results from this gives us a pretty good performance boost in some synthetic testing.

Tests were running jmeter with 100 threads in 100 loops.
```
Before:
Pull ms-images changeset: 3:45
Pull an invalid collection's changeset: 2:30

After:
Pull ms-images changeset: 1:40
Pull an invalid collection's changeset: 0:18
```

### DIFF
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -153,8 +153,8 @@ class Settings(BaseSettings):
         description="Sets the cache-control response header to max-age={value} for static content, like attachments. Default is 604800 (1 week)",
     )
     filter_refs_cache_size: int = Field(
-      500,
-      description="Number of filter_refs function results to cache. This filters git tags to a specific collection and is expensive to run per request."
+        500,
+        description="Number of filter_refs function results to cache. This filters git tags to a specific collection and is expensive to run per request.",
     )
 
 
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -126,7 +126,7 @@ logging.config.dictConfig(
 
 class Settings(BaseSettings):
     model_config = SettingsConfigDict(env_file=".env", extra="ignore", frozen=True)
-    git_repo_path: str | None = Field(None, description="Path to the Git repository")
+    git_repo_path: str = Field("", description="Path to the Git repository")
     self_contained: bool = Field(
         False,
         description="Whether to serve `attachments/` and `cert-chains/` endpoints.",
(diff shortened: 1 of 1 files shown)


git-reader/app.py
```diff
@@ -278,6 +278,28 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+@lru_cache(maxsize=500)
```
leplatrem: Where does this 500 come from? (number of possible collections + margin? maybe better as a config)

git-reader/app.py
```diff
@@ -278,6 +278,28 @@ def wrapper(*args, **kwargs):
...
+def filter_refs(
+    repo: pygit2.Repository,
+    bid: str,
+    cid: str,
```
leplatrem, line 289: Why don't we need any `cache_bust` here? How does the caching of references catch up the git repo update?

leplatrem (approved):
Good catch! I have doubt about caching logics though

git-reader/app.py
alexcottner: We are passing in the repo as a param, so anytime that ref changes (which should change whenever the last_modified timestamp changes) it would run the function again instead of returning a cached response.

git-reader/app.py
alexcottner: 500 is somewhat arbitrary, but yeah my thinking was this would leave a good amount of breathing room for our number of possible collections (preview + published).

I will make it configurable and default to 500, that seems like a good idea.

git-reader/app.py
alexcottner: Tested again to confirm this works as expected. The `get_repo` method's cache busting is working as expected, which means this is working properly.

leplatrem (approved):

git-reader/app.py
leplatrem: We could add a comment for future readers, because this is really not obvious at first look

git-reader/app.py
leplatrem: 👌

PR #1437, opened 2026-07-23
git-reader: gitupdate: pass LFS config via env vars

leplatrem:


git-reader/run.sh
```diff
@@ -30,6 +30,17 @@ cmd_gitupdate() {
     start_total=$(date +%s)
     log "Repository path is $repo_path"
 
+    # Apply LFS tuning from environment variables. If a variable is not set,
+    # leave the corresponding Git config untouched.
+    if [ -n "${LFS_CONCURRENT_TRANSFERS:-}" ]; then
+        log "Setting lfs.concurrenttransfers to ${LFS_CONCURRENT_TRANSFERS}."
+        git config --global lfs.concurrenttransfers "${LFS_CONCURRENT_TRANSFERS}"
+    fi
+    if [ -n "${LFS_FETCH_EXCLUDE:-}" ]; then
```
alexcottner, line 40: How much work would it be to abstract this a little bit more so the param could be just a list of buckets/collections?

alexcottner (approved):

PR #1473, opened 2026-08-13
RMST-497: Mitigate A-S client issue with tombstones

leplatrem:
https://mozilla-hub.atlassian.net/browse/RMST-497

- [Client specs](https://remote-settings.readthedocs.io/en/latest/client-specifications.html#local-state)
- [A-S client source](https://github.com/mozilla/application-services/blob/490bfb36c087698113502ce381dcfbb9aca7ec5c/components/remote_settings/src/client.rs#L819) does not have default on `last_modified` field
- [Desktop client](https://searchfox.org/firefox-main/rev/6cee80f87168d5aabd9e343099e2a8e5535e5320/services/settings/Database.sys.mjs#156)
- [RS team original rust client](https://github.com/mozilla-services/remote-settings-client/blob/2bc16799e475c71e6179a49536bf17b7554aea37/src/client.rs#L820-L821)

### DIFF
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -430,7 +430,8 @@ class GitService:
             # Deleted records are shown as tombstones.
             # Note: we set an arbitrary `last_modified` value as a
             # mitigation solution to A-S clients expecting it
-            # although it is not needed and not part of specifications.
+            # although it is not needed and not part of specifications
+            # (v1/ API had the field but was never officially mentioned).
             for rid in old_records_by_id.keys():
                 filtered[rid] = {"id": rid, "deleted": True, "last_modified": 0}
             records_by_id = filtered
--- a/git-reader/app.py
+++ b/git-reader/app.py
@@ -428,9 +428,11 @@ class GitService:
                 elif old_record != record:
                     filtered[rid] = record
             # Deleted records are shown as tombstones.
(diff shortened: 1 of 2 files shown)


git-reader/app.py
```diff
@@ -428,9 +428,12 @@ def get_collection_changeset(
...
                     filtered[rid] = record
             # Deleted records are shown as tombstones.
-            # Note: we don't have `last_modified` but clients don't need it.
+            # Note: we set an arbitrary `last_modified` value as a
+            # mitigation solution to A-S clients expecting it
+            # although it is not needed and not part of specifications
+            # (v1/ API had the field but was never officially mentioned).
             for rid in old_records_by_id.keys():
-                filtered[rid] = {"id": rid, "deleted": True}
+                filtered[rid] = {"id": rid, "deleted": True, "last_modified": 0}
```
alexcottner, line 436: Or should we put in the provided `_since` value instead of 0? I think we're mainly trying to make the deserialization happy.

alexcottner (approved):

git-reader/app.py
leplatrem: What value would this bring? 
I like `0` because it shows that we're not doing anything smart here

PR #1477, opened 2026-08-14
RMST-502: add ADR with possible solutions for missing tombstones

leplatrem:
What we had in this ticket:
https://mozilla-hub.atlassian.net/browse/RMST-502
in an ADR form...

PR #1546, opened 2026-09-07
git-export: RMST-502: add tombstones to ledger files

leplatrem:
See #1477 

First step for RMST-502: deploy this new git-export job that will store tombstones in ledgers files (see #1538).  Let it run a few times, stop it, run the backfill of #1545, restart the cronjob. 
At this point existing readers should work as usual. Then we can deploy the reader version of #1539 that reads from the ledger files.

### DIFF
--- a/cronjobs/src/commands/git_export.py
+++ b/cronjobs/src/commands/git_export.py
@@ -5,7 +5,7 @@ import os
 import traceback
 import urllib
 import urllib.parse
-from typing import Any, Iterable
+from typing import Any, Iterable, cast
 
 import kinto_http
 import pygit2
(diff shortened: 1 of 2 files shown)


cronjobs/src/commands/git_export.py
```diff
@@ -497,29 +516,75 @@ def process_attachments(
...
+        merged = known | entries
+        if merged == known:
+            # Nothing new: a full sync repeats every tombstone of the collection.
+            continue
+
+        ledger_files.append(
+            (
+                path,
+                "".join(
+                    f"{timestamp}{LEDGER_TIMESTAMP_SEPARATOR}{rid}\n"
```
leplatrem, line 558:
```suggestion
                "\n".join(
                    f"{timestamp}{LEDGER_TIMESTAMP_SEPARATOR}{rid}"
```

does not add a newline on files with a single entry...

alexcottner (approved):
Approved, I agree with your one suggestion though

leplatrem:
>  I agree with your one suggestion though

Me too
But as I was saying, for a single entry, we have no end of line

